PR #2, opened 2016-08-25
Ensure default make target is non-mutative

meatballhat:
and use an explicit terraform plan file to ensure zero drift

### DIFF
--- a/aws-staging/Makefile
+++ b/aws-staging/Makefile
@@ -1,16 +1,22 @@
+.PHONY: hello
+hello:
+	@echo "Hello there, human.  Would you like to:"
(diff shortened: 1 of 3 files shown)


aws-staging/Makefile
```diff
@@ -1,16 +1,22 @@
...
 apply: ensure-configured
 	terraform get
-	terraform apply
+	terraform apply $(PWD)/.terraform.plan
 
 .PHONY: plan
 plan: ensure-configured
 	terraform get
-	terraform plan
+	terraform plan -module-depth=-1 -out=$(PWD)/.terraform.plan
```
henrikhodne: `-module-depth=-1` is the default

PR #5, opened 2016-09-03
Starting work on refreshing AWS workers

meatballhat:
to work with travis-ci/worker


meatballhat:
I'm moving forward with this as-is given that AWS was in a mostly-non-working state before I started and the changes here are for both AWS and GCE.

PR #18, opened 2016-09-14
Plug cyclist into AWS

meatballhat:


meatballhat:
This is working enough in aws-staging that most of my changes have been happening in cyclist releases, so mergy mergy!

PR #21, opened 2016-09-16
Commenting-out GCE disk types in order to get staging-1 up

meatballhat:
as I was getting this error during a fresh `apply` for all `google_compute_instance` resources:

```
Error loading disk type 'pd-ssd': googleapi: got HTTP response code 404 with body: Not Found
```


meatballhat:
No longer needed yey :tada:

PR #25, opened 2016-09-21
Reduce chance of error in env name specification

meatballhat:
by using generated values based on env directory name


meatballhat:
@henrikhodne r?

meatballhat:
@henrikhodne yes, definitely, I would love to abstract this more!

PR #30, opened 2016-09-26
Clean up route tables and NAT for quay

meatballhat:
Targeting the now-very-much-overgrown #28


meatballhat:
@henrikhodne r?

meatballhat:
@igorwwwwwwwwwwwwwwwwwwww r?

aws-shared-1/main.tf
```diff
@@ -8,30 +13,66 @@ variable "workers_org_subnet_1e" { default = "10.10.6.0/24" }
...
+    protocol = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+}
+
+resource "aws_instance" "nat_quay" {
+  ami = "${var.nat_ami}"
+  instance_type = "c3.2xlarge"
+  source_dest_check = false
+  subnet_id = "${module.aws_az_1b.public_subnet_id}"
```
meatballhat: I figured pinning to az 1b was acceptable :shrug:

bin/generate-aws-route-table-public
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+main() {
+  cat <<EOF
+# WARNING: this file is generated!
```
henrikhodne: Would it be useful to also specify that it's this script that generated the file?

meatballhat:
I'm not 100% sure this will work, either, and given that I'm creating a new main route table association, I think this means I need to backfill some more of the default routes such as `10.0.0.0/8` local.

If dropping the quay NAT is an option, I'm all for it.

meatballhat:
Oh look! https://www.terraform.io/docs/providers/aws/r/route_table.html

> Note that the default route, mapping the VPC's CIDR block to "local", is created implicitly and cannot be specified.

PR #32, opened 2016-09-27
Cleaning up names and vars and exported bits for clarity

meatballhat:
plus reducing ASG mins to 0 to allow for faster testing cycles


meatballhat:
@emdantrim r?

aws-precise-staging-1/main.tf
```diff
@@ -44,14 +45,15 @@ module "aws_asg_org" {
...
+  env_short = "${var.env_short}"
   heroku_org = "${var.aws_heroku_org}"
   index = "${var.index}"
   security_groups = "${module.aws_az_1b.workers_org_security_group_id},${module.aws_az_1e.workers_org_security_group_id}"
   site = "org"
   syslog_address = "${var.syslog_address}"
   worker_ami = "${var.worker_ami}"
   worker_asg_max_size = "3"
-  worker_asg_min_size = "1"
+  worker_asg_min_size = "0"
```
emdantrim: intentional, I guess?

emdantrim: rather, will this make it so that a whole instance needs to start up before a staging build can be run?

aws-precise-staging-1/main.tf
```diff
@@ -80,14 +82,15 @@ module "aws_asg_com" {
...
+  env_short = "${var.env_short}"
   heroku_org = "${var.aws_heroku_org}"
   index = "${var.index}"
   security_groups = "${module.aws_az_1b.workers_com_security_group_id},${module.aws_az_1e.workers_com_security_group_id}"
   site = "com"
   syslog_address = "${var.syslog_address}"
   worker_ami = "${var.worker_ami}"
   worker_asg_max_size = "1"
-  worker_asg_min_size = "1"
+  worker_asg_min_size = "0"
```
emdantrim: same question as before

aws-shared-1/main.tf
```diff
@@ -2,22 +2,23 @@ variable "bastion_ami" { default = "ami-53d4a344" }
 variable "env" { default = "shared" }
 variable "index" { default = "1" }
 variable "nat_ami" { default = "ami-12c5b205" }
-variable "public_subnet_1b" { default = "10.10.1.0/24" }
-variable "public_subnet_1e" { default = "10.10.4.0/24" }
+variable "public_subnet_1b_cidr" { default = "10.10.1.0/24" }
```
emdantrim: for clarification -- we're putting "cidr" instead of just "subnet" to be 100% explicit that we're not talking about the subnet ID, right?

meatballhat: Yep!

aws-precise-staging-1/main.tf
meatballhat: In reality the ASG will have at least one instance spin up automatically because of the scaling rules, but having a minimum of 0 allows us to set the desired capacity to 0 without modifying the ASG, which is something we need to do when tearing down the entire VPC.

PR #33, opened 2016-09-27
Add route53 records for worker NATs

meatballhat:


meatballhat:
@emdantrim r?

aws-shared-1/main.tf
```diff
@@ -97,6 +97,28 @@ module "aws_az_1e" {
   workers_org_subnet_cidr = "${var.workers_org_subnet_1e_cidr}"
 }
 
+resource "aws_route53_record" "workers_org_nat" {
+  zone_id = "${var.travisci_net_external_zone_id}"
+  name = "workers-nat-org-${var.env}-${var.index}.aws-us-east-1.travisci.net"
+  type = "A"
+  ttl = "300"
+  records = [
```
emdantrim: If we have >1 AZ defined but each AZ's infrastructure is discrete, how can we be sure that we get any particular AZ at any time? or, alternatively, do we not care?

would we want to have a separate DNS entry for each AZ as well as one overarching DNS entry for if you wanted to hit Any Ol' AZ (which would theoretically be this record)?

meatballhat: I'm guessing by the code from which I'm copying that we mostly don't care, but I also like the idea of pre-AZ records in addition to these.

emdantrim: only reason I can think of that we'd be concerned about having that amount of granularity is if something was going wrong in one AZ but not in another -- but if we were to want to get into a specific bastion host, it'd be a bit of a treasure hunt to find out which IP is which host.

also, does route53 do reverse dns as well? b/c if that's the case then we could at least depend on reverse dns to be able to get the AZ out of the respective IPs.

meatballhat: For the case of bastions, we only specify per-AZ records :tada:  We'd have to specifically add a `PTR`/`.arpa` record for reverse dns iiuc.

meatballhat: I'm adding per-AZ records for NATs now

emdantrim: ah right bastions are separate from this.

sounds rad, then!

meatballhat:
:tada: yey

PR #37, opened 2016-09-29
Generate cyclist tokens

meatballhat:
since we should never need to have a record of them outside of terraform state or heroku config.

**NOTE** this diff includes re ordering the module declarations so that "com" comes before "org"

*update*: By default, the `random_id` resource will persist the result in the terraform state, and because we have no triggers defined it will not change unless we explicitly destroy/rebuild.  This can most easily be done with the `taint` command, e.g.:

``` bash
terraform taint random_id.cyclist_token_com
# => The resource random_id.cyclist_token_com in the module root has been marked as tainted! 
```


meatballhat:
@emdantrim r?

PR #39, opened 2016-09-30
Generate S3 and RabbitMQ creds and perms where possible

meatballhat:
plus inject some other creds from their sources of truth.

- [x] test RabbitMQ bits against a staging environment
- [x] verify S3 policies are sufficient for build caching


meatballhat:
@emdantrim r?

terraform-common.mk
```diff
@@ -0,0 +1,41 @@
+ENV_NAME := $(notdir $(shell cd $(PWD) && pwd))
```
Lyoness: what is the purpose of moving this from terraform.mk and doing `+include $(shell git rev-parse --show-toplevel)/terraform-common.mk` in original file?

aws-production-2/main.tf
```diff
@@ -103,9 +136,16 @@ module "aws_asg_org" {
   worker_asg_scale_in_threshold = 16
   worker_asg_scale_out_qty = 2
   worker_asg_scale_out_threshold = 8
+  worker_cache_bucket = "${var.worker_org_cache_bucket}"
```
Lyoness: are the cache bucket's the beginnings of moving off of quay.io for ASGs?  Or just holding state within AWS?

aws-production-2/main.tf
```diff
@@ -103,9 +136,16 @@ module "aws_asg_org" {
...
+### ${path.module}/config/worker-org-local.env
+${file("${path.module}/config/worker-org-local.env")}
+### ${path.module}/config/worker-org.env
+${file("${path.module}/config/worker-org.env")}
+### ${path.module}/worker.env
 ${file("${path.module}/worker.env")}
+
+export TRAVIS_WORKER_AMQP_URI=${module.rabbitmq_worker_config_org.uri}
 EOF
   worker_docker_image_android = "quay.io/travisci/ci-amethyst:packer-1473386113"
```
Lyoness: oh no wait, here do we need quay.io to hold the larger `universe` AMIs?

aws.mk
```diff
@@ -1,18 +1,43 @@
-include $(shell git rev-parse --show-toplevel)/terraform.mk
```
Lyoness: ah ok. so multiple makefiles are being parsed from `terraform.mk` as the top source of ultimate configurable truth :tm: ?

terraform-common.mk
Lyoness: ah I see the pattern, just trying to suss out the dep graph/structure

meatballhat: There were some bits I wanted to reuse, and others not, so I figured splitting it up was an OK solution :+1:

aws-production-2/main.tf
meatballhat: These cache buckets are used inside the jobs themselves, e.g. `cache: bundler`.  The bucket, key, and secret all used to be injected, and with this change only the bucket is injected while the IAM bits are generated in terraform land.

aws-production-2/main.tf
meatballhat: No change to anything about how we use quay happening here :smile_cat:, promise!

bin/env-to-envdir
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env ruby
```
emdantrim: @meatballhat this is an orphan :3

bin/generate-tfvars
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
emdantrim: @meatballhat very githubby!

solarce:
Since we're not going to be using the local Docker registry module `modules/aws_docker_registry/`, please add a quick `README.md` that clarifies it's not use right now, but here for when we need it. Thanks

modules/rabbitmq_user/main.tf
```diff
@@ -0,0 +1,37 @@
+variable "admin_password" {}
+variable "admin_username" {}
+variable "endpoint" {}
+variable "perm_configure" { default = ".*" }
+variable "perm_read" { default = ".*" }
+variable "perm_write" { default = ".*" }
```
emdantrim: yayyyy beginnings of granularity :beach_umbrella:

modules/aws_docker_registry/README.md
```diff
@@ -0,0 +1,7 @@
+# AWS docker registry
+
+**NOTE**: this is not currently being used in any VPC
```
emdantrim: :100:

Lyoness:
:taco: :tada:

PR #40, opened 2016-10-04
AWS Docker registry module

meatballhat:
with integration set aside for the moment


meatballhat:
Configuring the local registry as a pull-through cache while retaining security is going to require more work/research.  See: https://github.com/docker/distribution/issues/1486

PR #41, opened 2016-10-06
Cleaning up cloud-init for workers in AWS

meatballhat:
including:
- using cloud-init itself to do more heavy lifting
- removing the use of embedded variables in scripts
- maintaining hook scripts as separate files


meatballhat:
@Lyoness r?

modules/aws_asg/cloud-init.bash
```diff
@@ -30,8 +30,9 @@ main() {
   hosts_line="${instance_ipv4} ${instance_hostname} ${instance_hostname%%.*}"
   syslog_address="$(cat "${RUNDIR}/syslog-address")"
 
-  sed -i \
-    "s/___INSTANCE_ID___/${instance_id}/g" "${ETCDIR}/default/travis-worker"*
+  for envfile in "${ETCDIR}/default/travis-worker"*; do
```
Lyoness: :+1:

meatballhat:
@Lyoness sadly, I don't have a ton of confidence until I've used such changes in a full `terraform apply` and observed the hosts come up and behave as expected.  The addition of `bats` tests makes me a *wee* bit more confident, at least.

modules/aws_asg/cloud-init.bats
```diff
@@ -0,0 +1,116 @@
...
+EOF
+
+  cat >"${BATS_TMPDIR}/bin/mock" <<EOF
+#!/bin/bash
+echo "---> \$(basename \${0})" "\$@" >>${MOCKLOG}
+if [[ -f "${BATS_TMPDIR}/returns/\$(basename \${0})" ]]; then
+  cat "${BATS_TMPDIR}/returns/\$(basename \${0})"
+  exit 0
+fi
+echo "\${RANDOM}\${RANDOM}\${RANDOM}"
```
Lyoness: why do we test for random here?

Lyoness:
understood.  testing infra is hard.

modules/aws_asg/cloud-init.bats
meatballhat: I'm having the mocked-out executable echo a random string if there is no stub return value present, mostly so that every mocked-out executable *at least* has some unpredictable output so that I'm less likely to write brittle tests.

Lyoness:
ok!  Mergeosaurus!

PR #45, opened 2016-10-12
Ensure ec2 metadata API is unavailable on workers

meatballhat:


meatballhat:
@igorwwwwwwwwwwwwwwwwwwww r?

modules/aws_asg/cloud-init.bash
```diff
@@ -20,6 +20,8 @@ main() {
 
   service travis-worker stop || true
   service travis-worker start || true
+
+  iptables -A OUTPUT -d '169.254.169.254' -j DROP
```
Lyoness: ah. I had to look up difference of why `-drop` and not `-reject` :+1: TIL.

henrikhodne:
I don't think packets from Docker containers go through the `OUTPUT` chain in iptables, so I'm not sure if this will work.

tables_traverse

My understanding is that the packet goes PREROUTING -> FORWARD -> POSTROUTING (which is why our current fix for this is `iptables -t nat -I PREROUTING -p tcp -d 169.254.169.254 --dport 80 -j DNAT --to-destination 192.0.2.1`).

henrikhodne:
Something like `iptables -t nat -I PREROUTING -p tcp -d 169.254.169.254 -j DROP` might also work, but we should test to make sure that's the case.

PR #51, opened 2016-10-13
Add make targets to help with docker image update visibility

meatballhat:


meatballhat:
This isn't quite right...

PR #55, opened 2016-11-01
Some cleanups for gce-staging-1

meatballhat:
including:
- using correct keychain env name
- using correct gce JSON key


terraform-common.mk
```diff
@@ -1,6 +1,7 @@
 ENV_NAME := $(notdir $(shell cd $(PWD) && pwd))
 ENV_SHORT ?= $(word 2,$(subst -, ,$(ENV_NAME)))
 INFRA ?= $(word 1,$(subst -, ,$(ENV_NAME)))
+ENV_TAIL ?= $(subst $(INFRA)-,,$(ENV_NAME))
```
solarce: What is this for?

meatballhat: That's used in the `trvs` commands as the "environment" since we have a finer-grained separation of terraform infras than simply staging/production.

solarce: Got it, thanks 👍

PR #57, opened 2016-11-03
Set aws-precise-staging-1 to use same worker ami as aws-staging-1

meatballhat:


meatballhat:
@Lyoness The AMIs come out of packer-templates.  Ideally, we should be able to replace the hardcoded AMI IDs with a data source query for the latest matching certain criteria, as is being done in the `./aws-shared-*/main.tf` for the bastion AMI.

PR #61, opened 2016-11-13
Re-enable and break up script that handles unregister_netdevice

meatballhat:


meatballhat:
This is working correctly in staging :+1:

meatballhat:
Confirmed working correctly on an elder production worker instance :+1:

igorwwwwwwwwwwwwwwwwwwww:
@meatballhat any reason why we cannot get rid of this completely once we've upgraded to mainline?

meatballhat:
@igorwwwwwwwwwwwwwwwwwwww I consider this a reasonable safeguard regardless of the kernel version we're on.  If the error ever re-surfaces, for whatever reason, we'll be prepared.

igorwwwwwwwwwwwwwwwwwwww:
Okay, I'm fine with keeping it around. :)

PR #62, opened 2016-11-14
Expand production to include us-east-1{a,c}

meatballhat:


meatballhat:
This is currently live in production, fwiw :+1:

PR #64, opened 2016-11-15
Bump the production worker AMI to use upgraded kernel

meatballhat:


meatballhat:
I'm merging this now, and will apply it via terraform per [the schedule shown here](https://docs.travis-ci.com/user/build-environment-updates/2016-11-15/).

PR #66, opened 2016-11-16
Update all images for GCE staging

meatballhat:


meatballhat:
This has already been applied.

PR #79, opened 2016-11-29
Disallow internal traffic in worker subnet

igorwwwwwwwwwwwwwwwwwwww:


modules/gce_project/network.tf
```diff
@@ -62,23 +62,3 @@ resource "google_compute_firewall" "allow_ssh" {
...
-
-resource "google_compute_firewall" "allow_internal" {
-  name = "allow-internal"
-  network = "${google_compute_network.main.name}"
-  source_ranges = ["10.10.0.0/16"]
-
-  project = "${var.project}"
-
-  allow {
-    protocol = "icmp"
```
henrikhodne: Keeping ICMP might be a good idea.

henrikhodne: Debugging, and there are several parts of IP that doesn't work very well without ICMP enabled (I don't remember them off the top of my head, but I'll see if I can dig up a link tomorrow).

igorwwwwwwwwwwwwwwwwwwww: @henrikhodne fyi, icmp from outside is allowed.

PR #82, opened 2016-12-02
Rework GCE staging in preparation for NAT

igorwwwwwwwwwwwwwwwwwwww:
This is a relatively large patch in terms of functionality, the changes so far:

* update to newer gce worker image (from packer)
* configure syslog for the worker instance (gives us kernel logs)
* merge `workers_com` and `workers_org` subnets to a single `workers` one
* create two separate `build_com` and `build_org` subnets for build instances
* disallow traffic between build instances
* allow traffic from worker to build instance
* use yml based cloud-config (just like on aws, this fixes problems with order of execution)


meatballhat:
I realize it will result in more diff churn, but I'd like to move toward the style suggested by Hashifolk where each terraform directory includes a `main.tf` with variables at the top and outputs at the bottom.

igorwwwwwwwwwwwwwwwwwwww:
Since this is what we are running on staging, I am going to go ahead and merge this PR.

PR #85, opened 2016-12-09
[gce-staging-1] Increase size of subnets to match what we are going for in production

igorwwwwwwwwwwwwwwwwwwww:


igorwwwwwwwwwwwwwwwwwwww:
This has been applied to staging.

henrikhodne:
Would it be good to use the same ranges as in production? `10.10.8.0/22` for .org and `10.10.12.0/22` for .com?

igorwwwwwwwwwwwwwwwwwwww:
can do, what should we do with the `public` and `workers` subnets?

henrikhodne:
`10.10.0.0/22` and `10.10.4.0/22`?

igorwwwwwwwwwwwwwwwwwwww:
done!

igorwwwwwwwwwwwwwwwwwwww:
also applied now.

PR #88, opened 2017-01-11
add pprof_port to aws worker config

igorwwwwwwwwwwwwwwwwwwww:
Enabling pprof so that we can run go profiling tools and such against worker :)


igorwwwwwwwwwwwwwwwwwwww:
@meatballhat yep, those are currently still managed via keychain. (PRs already done and merged)

PR #93, opened 2017-01-16
add gce-staging-2

igorwwwwwwwwwwwwwwwwwwww:
to be used for nat load testing and other things that we do not want to impact staging-1

### DIFF
--- a/gce-staging-2/main.tf
+++ b/gce-staging-2/main.tf
@@ -1,4 +1,4 @@
-variable "env" { default = "staging-2" }
+variable "env" { default = "staging" }
 variable "gce_bastion_image" { default = "eco-emissary-99515/bastion-1478778272" }
 variable "gce_gcloud_zone" {}
 variable "gce_heroku_org" {}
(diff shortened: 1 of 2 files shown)


gce-staging-2/main.tf
```diff
@@ -0,0 +1,51 @@
+variable "env" { default = "staging-2" }
```
meatballhat: This should be `default = "staging"`

gce-staging-2/main.tf
```diff
@@ -0,0 +1,51 @@
...
+variable "gce_bastion_image" { default = "eco-emissary-99515/bastion-1478778272" }
+variable "gce_gcloud_zone" {}
+variable "gce_heroku_org" {}
+variable "gce_nat_image" { default = "eco-emissary-99515/nat-1478778271" }
+variable "gce_hashistack_server_image" { default = "eco-emissary-99515/hashistack-server-1480351044" }
+variable "gce_worker_image" { default = "eco-emissary-99515/travis-worker-1480649763" }
+variable "github_users" {}
+variable "index" { default = 1 }
```
meatballhat: This should be `default = 2`

igorwwwwwwwwwwwwwwwwwwww: ah yes, that is a nicer way of doing it. thanks! :)

igorwwwwwwwwwwwwwwwwwwww:
@meatballhat applied suggested change!

PR #97, opened 2017-01-24
Generate tfvars for better error handling

### LINKED ISSUE #74: generate-tfvars not re-generated on empty file
If an error occurs during `bin/generate-tfvars`, an empty file is written and re-generating is not re-attempted.

This requires it to be deleted manually, which is not very nice.

meatballhat:
Closes #74

### DIFF
--- a/bin/generate-tfvars
+++ b/bin/generate-tfvars
@@ -5,9 +5,25 @@ set -o errexit
 main() {
   local top
   top="$(git rev-parse --show-toplevel)"
+  local out="${1}"
+  if [[ "${out}" ]]; then
+    exec 1<>"${out}"
+  fi
 
-  "${top}/bin/generate-github-ssh-users"
-  "${top}/bin/generate-latest-docker-image-tags"
+  if ! "${top}/bin/generate-github-ssh-users"; then
+    __cleanup "${out}"
+    exit 1
+  fi
+
+  if ! "${top}/bin/generate-latest-docker-image-tags"; then
+    __cleanup "${out}"
+  fi
+}
+
+__cleanup() {
+  if [[ "${1}" && -f "${1}" ]]; then
+    rm -f "${1}"
+  fi
 }
 
 main "$@"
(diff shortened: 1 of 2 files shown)


bin/generate-tfvars
```diff
@@ -5,9 +5,25 @@ set -o errexit
 main() {
   local top
   top="$(git rev-parse --show-toplevel)"
+  local out="${1}"
+  if [[ "${out}" ]]; then
+    exec 1<>"${out}"
```
henrikhodne: The `<>` can be a `>` here since I don't think we need to read from the file?

bin/generate-tfvars
```diff
@@ -5,9 +5,25 @@ set -o errexit
...
 
-  "${top}/bin/generate-github-ssh-users"
-  "${top}/bin/generate-latest-docker-image-tags"
+  if ! "${top}/bin/generate-github-ssh-users"; then
+    __cleanup "${out}"
+    exit 1
+  fi
+
+  if ! "${top}/bin/generate-latest-docker-image-tags"; then
+    __cleanup "${out}"
```
henrikhodne: Should we `exit 1` here too to have the binary as a whole still fail?

meatballhat: looks like the change wasn't detected, but this has been addressed

terraform-common.mk
```diff
@@ -47,4 +47,4 @@ graph:
 	terraform graph -draw-cycles | dot -Tpng > graph.png
 
 $(ENV_NAME).tfvars:
-	$(TOP)/bin/generate-tfvars >$@
+	$(TOP)/bin/generate-tfvars $@
```
henrikhodne: Not a bug that was introduced in this PR, but `$@` should be `"$@"`, otherwise paths with spaced in them will break.

meatballhat: @henrikhodne this var is resolved by `make` rather than `bash`, though.  It's `$(ENV_NAME).tfvars`.

meatballhat: Or did you mean it should be quoted to prevent issues with `$(ENV_NAME).tfvars` having spaces?

henrikhodne: Oh, good point, I missed that this was Make and not bash, :man_facepalming:. Not sure how Make handles `$@` and spaces and all, but since I don't think `$(ENV_NAME)` will have spaces in it(?) then I don't think it's really much of an issue.

PR #98, opened 2017-02-02
Switch to cloud formation ASGs

meatballhat:
for automatic rollout of new launch configurations

See: https://github.com/hashicorp/terraform/issues/1552#issuecomment-191847434

### DIFF
--- a/bin/generate-latest-docker-image-tags
+++ b/bin/generate-latest-docker-image-tags
@@ -33,6 +33,10 @@ def main
 end
 
 def latest_docker_image_tag(image)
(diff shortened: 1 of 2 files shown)


igorwwwwwwwwwwwwwwwwwwww:
Would like more details on what cloudformation gives us, how it differs from the existing setup, and if there are any stability concerns. And perhaps why this is relevant for us.

Thanks!

meatballhat:
@igorwwwwwwwwwwwwwwwwwwww tl;dr - updating a launch configuration automatically propagates.See the linked terraform thread for oodles of details.

PR #112, opened 2017-02-14
macstadium-shared-1 makefile fixes

henrikhodne:
This should speed up the `make plan` and `make apply` runs by speeding up all the config generation / only running it once.

### DIFF
--- a/macstadium-shared-1/Makefile
+++ b/macstadium-shared-1/Makefile
@@ -36,51 +36,51 @@ $(CONFIG_FILES):
 	cp -v $$TRAVIS_KEYCHAIN_DIR/travis-keychain/macstadium/travis-vm-ssh-key config/
 	trvs generate-config -p JUPITER_BRAIN -f env jupiter-brain production-1 \
 		| sed 's/^/export /' >config/jupiter-brain-prod-org-env
-	trvs generate-config -p JUPITER_BRAIN -f env jupiter-brain staging-1 \
+	trvs generate-config -n -p JUPITER_BRAIN -f env jupiter-brain staging-1 \
 		| sed 's/^/export /' >config/jupiter-brain-staging-org-env
-	trvs generate-config -p JUPITER_BRAIN -f env jupiter-brain custom-1 \
+	trvs generate-config -n -p JUPITER_BRAIN -f env jupiter-brain custom-1 \
 		| sed 's/^/export /' >config/jupiter-brain-custom-1-env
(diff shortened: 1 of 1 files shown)


macstadium-shared-1/Makefile
```diff
@@ -18,51 +36,51 @@ $(CONFIG_FILES):
...
+	trvs generate-config -n --pro -p TRAVIS_WORKER -f env macstadium-workers staging-1-2 \
 		| sed 's/^/export /' >config/travis-worker-com-staging-2
-	trvs generate-config --pro -p TRAVIS_WORKER -f env macstadium-workers production-1-1 \
+	trvs generate-config -n --pro -p TRAVIS_WORKER -f env macstadium-workers production-1-1 \
 		| sed 's/^/export /' >config/travis-worker-com-prod-1
-	trvs generate-config --pro -p TRAVIS_WORKER -f env macstadium-workers production-1-2 \
+	trvs generate-config -n --pro -p TRAVIS_WORKER -f env macstadium-workers production-1-2 \
 		| sed 's/^/export /' >config/travis-worker-com-prod-2
-	trvs generate-config -p COLLECTD_VSPHERE -f env collectd-vsphere common-1 \
+	trvs generate-config -n -p COLLECTD_VSPHERE -f env collectd-vsphere common-1 \
```
igorwwwwwwwwwwwwwwwwwwww: IMO the keychain update behaviour is something we generally want. For testing out different keychain branches I would rather go with the `TRAVIS_SKIP_KEYCHAIN_UPDATE=true` env var.

henrikhodne: I'm only skipping it on the generates _after_ the first one, so both keychains get updated once, but then not again for the remaining files.

henrikhodne: I think that's good for two reasons:

1. It's faster because it's fewer `git pulls`
1. It makes sure that we only have one commit of keychain for all the configs, instead of currently when it's possible that someone pushes something in the middle of generating all the configs, and we'll have one version for some commits and another for other configs.

(Also, there's nothing ensuring that we're getting the `master` branch, this still runs fine if you have a branch of keychain checked out, it just makes sure that you have everything that's been pushed to GitHub. You don't even need to push your latest changes, as long as it's committed and the branch exists on the remote)

igorwwwwwwwwwwwwwwwwwwww: ah, missed that detail :) in this case I think it's fine 👍

PR #114, opened 2017-02-14
fetch as much config as possible from the keychain

igorwwwwwwwwwwwwwwwwwwww:
Refs https://github.com/travis-pro/travis-keychain/pull/216.

That means `make plan` should work off the bat, as long as `TF_VAR_ssh_user` is set (as per `.exmaple.env`)!

### DIFF
--- a/macstadium-shared-1/main.tf
+++ b/macstadium-shared-1/main.tf
@@ -23,9 +23,17 @@ variable "collectd_vsphere_collectd_network_user" {}
 variable "collectd_vsphere_collectd_network_token" {}
 variable "fw_ip" {}
 variable "fw_snmp_community" {}
+variable "vsphere_user" {}
+variable "vsphere_password" {}
+variable "vsphere_server" {}
 
 provider "aws" {}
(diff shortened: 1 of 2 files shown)


macstadium-shared-1/main.tf
```diff
@@ -23,9 +23,17 @@ variable "collectd_vsphere_collectd_network_user" {}
...
+variable "vsphere_password" {}
+variable "vsphere_server" {}
 
 provider "aws" {}
-provider "vsphere" {}
+provider "vsphere" {
+  user           = "${var.vsphere_user}"
+  password       = "${var.vsphere_password}"
+  vsphere_server = "${var.vsphere_server}"
+  allow_unverified_ssl = true
```
henrikhodne: Should this also be configurable?

igorwwwwwwwwwwwwwwwwwwww: I'll leave it out for now, we can make it configurable if we get into a situation where some require ssl. Hope that's ok!

PR #116, opened 2017-02-14
macstadium-shared-2: terraform config for second pod based on macstadium-shared-1

igorwwwwwwwwwwwwwwwwwwww:
* diff: https://gist.github.com/igorwwwwwwwwwwwwwwwwwwww/cec2265c319bcca4ec995ce72cfbca98
* config: https://github.com/travis-pro/travis-keychain/pull/217

### DIFF
--- a/macstadium-shared-2/Makefile
+++ b/macstadium-shared-2/Makefile
@@ -1,0 +1,88 @@
+include $(shell git rev-parse --show-toplevel)/terraform.mk
+
+.PHONY: default
+default: hello
+
+CONFIG_FILES := \
+		config/jupiter-brain-prod-org-env \
+		config/jupiter-brain-staging-org-env \
+		config/jupiter-brain-custom-1-env \
+		config/jupiter-brain-custom-2-env \
+		config/jupiter-brain-custom-3-env \
+		config/travis-worker-org-staging-1 \
+		config/travis-worker-org-staging-2 \
+		config/travis-worker-org-prod-1 \
+		config/travis-worker-org-prod-2 \
+		config/travis-worker-custom-1 \
+		config/travis-worker-custom-2 \
+		config/travis-worker-custom-3 \
+		config/vsphere-janitor-prod-com \
+		config/vsphere-janitor-staging-com \
+		config/vsphere-janitor-custom-1 \
+		config/vsphere-janitor-custom-2 \
+		config/vsphere-janitor-custom-3 \
+		config/jupiter-brain-prod-com-env \
+		config/jupiter-brain-staging-com-env \
+		config/travis-worker-com-staging-1 \
+		config/travis-worker-com-prod-1 \
+		config/collectd-vsphere-common-1 \
+		config/travis-vm-ssh-key
+
+.PHONY: .config
+.config: $(CONFIG_FILES) $(ENV_NAME).tfvars
+
+$(CONFIG_FILES):
+	mkdir -p config
+	cp -v $$TRAVIS_KEYCHAIN_DIR/travis-keychain/macstadium/travis-vm-ssh-key config/
+	trvs generate-config -p JUPITER_BRAIN -f env jupiter-brain production-1 \
+		| sed 's/^/export /' >config/jupiter-brain-prod-org-env
+	trvs generate-config -n -p JUPITER_BRAIN -f env jupiter-brain staging-1 \
+		| sed 's/^/export /' >config/jupiter-brain-staging-org-env
+	trvs generate-config -n -p JUPITER_BRAIN -f env jupiter-brain custom-1 \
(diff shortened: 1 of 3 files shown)


macstadium-shared-2/Makefile
```diff
@@ -0,0 +1,61 @@
...
+		| sed 's/^/export /' >config/jupiter-brain-staging-com-env
+	trvs generate-config -n --pro -p TRAVIS_WORKER -f env macstadium-workers staging-1-1 \
+		| sed 's/^/export /' >config/travis-worker-com-staging-1
+	trvs generate-config -n --pro -p TRAVIS_WORKER -f env macstadium-workers staging-1-2 \
+		| sed 's/^/export /' >config/travis-worker-com-staging-2
+	trvs generate-config -n --pro -p TRAVIS_WORKER -f env macstadium-workers production-1-1 \
+		| sed 's/^/export /' >config/travis-worker-com-prod-1
+	trvs generate-config -n --pro -p TRAVIS_WORKER -f env macstadium-workers production-1-2 \
+		| sed 's/^/export /' >config/travis-worker-com-prod-2
+	trvs generate-config -n -p COLLECTD_VSPHERE -f env collectd-vsphere common-1 \
```
henrikhodne: This should be `common-2` instead of `common-1`

macstadium-shared-2/Makefile
```diff
@@ -0,0 +1,61 @@
...
+.config: $(CONFIG_FILES) $(ENV_NAME).tfvars
+
+$(CONFIG_FILES):
+	mkdir -p config
+	cp -v $$TRAVIS_KEYCHAIN_DIR/travis-keychain/macstadium/travis-vm-ssh-key config/
+	trvs generate-config -p JUPITER_BRAIN -f env jupiter-brain production-1 \
+		| sed 's/^/export /' >config/jupiter-brain-prod-org-env
+	trvs generate-config -n -p JUPITER_BRAIN -f env jupiter-brain staging-1 \
+		| sed 's/^/export /' >config/jupiter-brain-staging-org-env
+	trvs generate-config -n -p TRAVIS_WORKER -f env macstadium-workers staging-1-1 \
```
henrikhodne: This (and all the `1-1` and `1-2` etc.) below it should be `2-1`, the first number is the pod.

macstadium-shared-2/main.tf
```diff
@@ -0,0 +1,195 @@
...
+  ssh_ip_address = "${module.macstadium_infrastructure.wjb_ip}"
+  ssh_user = "${var.ssh_user}"
+  version = "${var.jupiter_brain_prod_version}"
+  config_path = "${path.module}/config/jupiter-brain-staging-com-env"
+  env = "com-staging"
+  index = "${var.index}"
+  port_suffix = 4
+}
+
+module "jupiter_brain_custom_1" {
```
henrikhodne: I think the custom workers and jupiter brains should be removed for now, since they are currently only in one pod.

macstadium-shared-2/Makefile
igorwwwwwwwwwwwwwwwwwwww: if the -1 is part of the pod, would it make sense to rename `common-1` to `common` in the filenames and such, similar to how it's done with the others?

macstadium-shared-2/Makefile
igorwwwwwwwwwwwwwwwwwwww: fixing!

macstadium-shared-2/main.tf
igorwwwwwwwwwwwwwwwwwwww: fixing!

igorwwwwwwwwwwwwwwwwwwww:
Updated!

macstadium-shared-2/main.tf
```diff
@@ -0,0 +1,153 @@
...
+module "collectd-vsphere-common-2" {
+  source = "../modules/collectd_vsphere"
+  host_id = "${module.macstadium_infrastructure.wjb_uuid}"
+  ssh_host = "${module.macstadium_infrastructure.wjb_ip}"
+  ssh_user = "${var.ssh_user}"
+  version = "${var.collectd_vsphere_version}"
+  config_path = "${path.module}/config/collectd-vsphere-common-2"
+  librato_email = "${var.librato_email}"
+  librato_token = "${var.librato_token}"
+  env = "common-2"
```
igorwwwwwwwwwwwwwwwwwwww: this looks strange, @henrikhodne confirm that this is what we want?

henrikhodne: That looks correct to me. `pod-2` (and `pod-1` in the other one) might be better now that we've"standardized" on the "pod" wording, but with the current wording everywhere this looks right.

PR #127, opened 2017-02-16
terraform fmt everything

henrikhodne:
This is from me running `terraform fmt` in the root directory of the repository, which is similar to `go fmt`.


henrikhodne:
Looks like `terraform fmt` is doing different things on my machine and on Travis, not sure why… will come back to this later.

henrikhodne:
The next version of Terraform (0.8.8 or 0.9) should have a fix for the issue I was seeing, planning on revisiting this once that is released.

emdantrim:
Closing this to re-run `terraform fmt` against master.

PR #128, opened 2017-02-16
macstadium: move vm ssh key upload to wjb tf file

henrikhodne:
This file only exists once per wjb, so having it in the worker module caused all the parallel worker modules to "fight" over the file as they were running, causing some odd race conditions.

### DIFF
--- a/modules/macstadium_infrastructure/wjb.tf
+++ b/modules/macstadium_infrastructure/wjb.tf
@@ -44,5 +44,30 @@ resource "vsphere_virtual_machine" "wjb" {
   }
 }
 
+resource "null_resource" "worker" {
+  triggers {
+    ssh_key_file_signature = "${sha256(file(var.vm_ssh_key_path))}"
+  }
+
+  connection {
+    host = "${vsphere_virtual_machine.wjb.network_interface.0.ipv4_address}"
+    user = "${var.ssh_user}"
+    agent = true
+  }
+
+  provisioner "file" {
+    source      = "${var.vm_ssh_key_path}"
+    destination = "/tmp/travis-vm-ssh-key"
+  }
+
+  provisioner "remote-exec" {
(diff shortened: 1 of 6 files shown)


modules/macstadium_infrastructure/wjb.tf
```diff
@@ -44,5 +44,30 @@ resource "vsphere_virtual_machine" "wjb" {
...
+
+  provisioner "file" {
+    source      = "${var.vm_ssh_key_path}"
+    destination = "/tmp/travis-vm-ssh-key"
+  }
+
+  provisioner "remote-exec" {
+    inline = [
+      "sudo mv /tmp/travis-vm-ssh-key /etc/travis-vm-ssh-key",
+      "sudo chown travis-worker:travis-worker /etc/travis-vm-ssh-key",
```
henrikhodne: Shoot, this is going to break on new provisions since that user doesn't exist yet :thinking:

modules/macstadium_infrastructure/wjb.tf
```diff
@@ -44,5 +44,32 @@ resource "vsphere_virtual_machine" "wjb" {
...
+  }
+
+  provisioner "file" {
+    source      = "${var.vm_ssh_key_path}"
+    destination = "/tmp/travis-vm-ssh-key"
+  }
+
+  provisioner "remote-exec" {
+    inline = [
+      "if ! getent passwd travis-worker >/dev/null; then sudo useradd -r -s /usr/bin/nologin travis-worker; fi"
```
henrikhodne: This has a race condition since other provisioners might create the user between `getent` and `useradd`, but the same code is used elsewhere so I'm not too worried. We could always add some `|| getent passd travis-worker >/dev/null` again after the `useradd` if it becomes a problem.

PR #136, opened 2017-03-07
jupiter_brain: use addition instead of string concatenation for ports

henrikhodne:
This allows more than 10 ports.


Lyoness:
Are we running into an issue where we need more than 10 ports?  Can you explain so that I can understand basics for my own understanding re: jupiter-brain networking internals?

henrikhodne:
> Are we running into an issue where we need more than 10 ports? Can you explain so that I can understand basics for my own understanding re: jupiter-brain networking internals?

Yeah, I added the 11th Jupiter Brain in #135. Up until then port_suffix was a single digit number, so we'd have ports like `8081` and `8088`, but with `port_suffix = 10`, we got port `80810` instead of `8090`, but this PR should fix that.

PR #139, opened 2017-03-08
Add .org production workers

henrikhodne:
This adds .org production Jupiter Brains and workers in both pods.

Also includes a small fix to add us-east-1 as the default region for AWS, since Terraform was complaining about no region being set when trying to apply things.


solarce:
This is for the beginning of Phase 3 in https://github.com/travis-pro/team-blue/issues/592

PR #143, opened 2017-03-15
Config fixes and updates for custom-2 and custom-3

solarce:
- Adding haproxy configs to `pod-2`
- Adding missing janitor configs for `pod-2`
- Fixing custom-3 ports for `pod-1` and `pod-2`


solarce:
This has been applied in production.

PR #144, opened 2017-03-17
Upgrade to Terraform 0.9

henrikhodne:
See https://www.terraform.io/docs/backends/legacy-0-8.html for upgrade guide for the remote state.

**Note**: I already did the upgrade for `macstadium-shared-1`, as I didn't realize that it would immediately update the remote state as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
     pushd "${TMPDIR}" ;
     curl -sSL
       -o terraform.zip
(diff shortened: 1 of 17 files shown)


henrikhodne:
> I feel like we should add a version check, probably in terraform-common.mk?, that verifies someone's local setup has the appropriate terraform version and exits if they don't, is a necessity. @meatballhat @henrikhodne, thoughts?

Terraform itself checks the version in the remote state and will error out if the remote state has a newer version than you have locally.

PR #148, opened 2017-04-11
Add gce-production-1 project

henrikhodne:
This is replacing an existing project, so some tweaking was necessary with the CIDR ranges to not conflicting with the existing networks. I made the variables have their existing values as defaults, so it should be backwards-compatible for production-6 and production-7.


henrikhodne:
For the record, these changes are already applied.

PR #155, opened 2017-04-20
Add custom-6 services

emdantrim:


macstadium-shared-1/main.tf
```diff
@@ -201,6 +201,18 @@ module "jupiter_brain_custom_5" {
...
+module "jupiter_brain_custom_6" {
+  source = "../modules/jupiter_brain_bluegreen"
+  host_id = "${module.macstadium_infrastructure.wjb_uuid}"
+  ssh_ip_address = "${module.macstadium_infrastructure.wjb_ip}"
+  ssh_user = "${var.ssh_user}"
+  version = "${var.jupiter_brain_custom-6_version}"
+  config_path = "${path.module}/config/jupiter-brain-custom-6-env"
+  env = "custom-6"
+  index = "${var.index}"
+  port_suffix = 9
```
henrikhodne: This conflicts with custom-5, I think. Looking at the ports below, I think it should be 11?

PR #167, opened 2017-06-06
macstadium-shared: change haproxy config var syntax to explicit list …

igorwwwwwwwwwwwwwwwwwwww:
…of maps

It was giving us this error:

```
1 error(s) occurred:

* module.haproxy.var.config: variable config in module haproxy should be type list, got map
```


igorwwwwwwwwwwwwwwwwwwww:
We suspect this may have changed between terraform versions.

PR #175, opened 2017-06-14
aws: switch from simple scaling to step scaling to allow faster scale out

igorwwwwwwwwwwwwwwwwwwww:
refs https://github.com/travis-pro/dev/issues/45

one of the issues we are running into is this scaling limitation ([source](http://docs.aws.amazon.com/autoscaling/latest/userguide/lifecycle-hooks.html)):

> When the instance enters the wait state, scaling actions due to simple scaling policies are suspended.

it turns out that there is a different type of scaling policy called step scaling, which appears not to be affected by that limitation. better yet, the "warmup" periods that can be specified appear to actually factor in slow boot and shutdown times into the scaling itself.

it looks like this might do _exactly_ what we want. will need some experimentation and testing to verify that though.

see also:

* https://docs.aws.amazon.com/autoscaling/latest/userguide/as-scale-based-on-demand.html#as-scaling-types
* https://www.terraform.io/docs/providers/aws/r/autoscaling_policy.html
* https://aws.amazon.com/blogs/aws/auto-scaling-update-new-scaling-policies-for-more-responsive-scaling/


modules/aws_asg/main.tf
```diff
@@ -265,11 +265,15 @@ resource "aws_autoscaling_group" "workers" {
...
 
 resource "aws_autoscaling_policy" "workers_remove_capacity" {
-  name                   = "${var.env}-${var.index}-workers-${var.site}-remove-capacity"
-  scaling_adjustment     = "${var.worker_asg_scale_in_qty}"
-  adjustment_type        = "ChangeInCapacity"
-  cooldown               = "${var.worker_asg_scale_in_cooldown}"
-  autoscaling_group_name = "${aws_autoscaling_group.workers.name}"
+  name                      = "${var.env}-${var.index}-workers-${var.site}-remove-capacity"
+  adjustment_type           = "ChangeInCapacity"
+  policy_type               = "StepScaling"
```
meatballhat: I'm confused about our current scaling policy being "simple" given the terraform docs about `policy_type` 😕 https://www.terraform.io/docs/providers/aws/r/appautoscaling_policy.html#policy_type which indicate we're using `StepScaling` whether or not we specify it.  Sure enough, source says: https://github.com/hashicorp/terraform/blob/5f24491ea6659c50883d85a063f902a22b8c2a50/builtin/providers/aws/resource_aws_appautoscaling_policy.go#L40-L44

modules/aws_asg/main.tf
```diff
@@ -265,11 +265,15 @@ resource "aws_autoscaling_group" "workers" {
...
+  autoscaling_group_name    = "${aws_autoscaling_group.workers.name}"
+  estimated_instance_warmup = "${var.worker_asg_scale_in_cooldown}"
+
+  step_adjustment {
+    scaling_adjustment = "${var.worker_asg_scale_in_qty}"
+  }
```
meatballhat: IIUC we need to specify multiple `step_adjustment` blocks in order to take advantage of the stepping, no?  ...which makes sense to me given the AWS blog example and https://www.terraform.io/docs/providers/aws/r/appautoscaling_policy.html#step_adjustment

igorwwwwwwwwwwwwwwwwwwww:
@meatballhat we are not using `aws_appautoscaling_policy` but `aws_autoscaling_policy`: https://www.terraform.io/docs/providers/aws/r/autoscaling_policy.html

Lyoness:
There's no viable testing solution for this change,  @igorwwwwwwwwwwwwwwwwwwww and I have brainstormed possible ideas/ solutions for this [here](https://travisci.slack.com/archives/C0305BDHP/p1502200581724504)

further discussion welcome.

soulshake:
Tested in https://github.com/travis-infrastructure/terraform-config/pull/222

soulshake:
Just pushed the new policies to production:

## Before
old-scaling-policies-org

## After
new-scaling-policies-org

PR #183, opened 2017-06-29
Run travis-worker on a tfw image

meatballhat:
including moving service config files here

I have applied this to:
- aws-staging-1
- gce-staging-1
- gce-staging-2


igorwwwwwwwwwwwwwwwwwwww:
I have a question! what is the difference between the worker image and the new tfw image?

meatballhat:
@igorwwwwwwwwwwwwwwwwwwww The worker image had a few layers in play, with travis-cookbooks, then packer-templates, then terraform-config all contributing parts to a whole.  The tfw image is meant as a more general purpose Ubuntu machine with Docker, with the only worker-specific code being added at cloud-init time via terraform-config.

PR #199, opened 2017-08-09
Add 'make check' to verify a few common setup requirements

soulshake:
After running into a couple of hangups getting to `make plan`, I thought it might be helpful to add a `make check` target to verify some required envvars and credentials.

Thoughts?

### DIFF
--- a/bin/check-credentials
+++ b/bin/check-credentials
@@ -6,38 +6,37 @@ die() {
 }
 
 stderr_echo() {
-  >&2 echo "$@"
+  echo >&2 "$@"
 }
 
 check_heroku() {
-    endpoints="
-        travis-pro-web-staging
+  endpoints="
+        travis-pro-staging
         travis-staging
     "
-    stderr_echo "Attempting to authenticate with Heroku..."
-    for endpoint in $endpoints; do
-        cmd="curl --silent \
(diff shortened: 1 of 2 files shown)


bin/check-credentials
```diff
@@ -0,0 +1,70 @@
...
+  stderr_echo "$@" && exit 1
+}
+
+stderr_echo() {
+  >&2 echo "$@"
+}
+
+check_heroku() {
+    endpoints="
+        travis-pro-web-staging
```
igorwwwwwwwwwwwwwwwwwwww: maybe change this to `travis-pro-staging`? that way the arbitrary apps we check follow some sort of pattern.

bogdanap:
Isn't it worth merging this straight into master rather than meat-worker-rollout?

igorwwwwwwwwwwwwwwwwwwww:
As per slack, since `bin/travis-worker-verify-config` is run as a local-exec via terraform, merging them may not be so straight forward. So nevermind that, I think we can move forward with this as-is!

@bogdanap I'd generally agree that targeting master and then merging master back into branches is preferable. 👍 should be better once `meat-worker-rollout` lands in master.

soulshake:
Updated to merge directly to master.

PR #203, opened 2017-08-09
Bumping max ASG size in trusty to match precise

kdaniels:


Lyoness:
Addresses: https://github.com/travis-pro/team-blue/issues/724

PR #208, opened 2017-08-22
Make `make check` check more things

soulshake:

- Some issues can cause silent or unclear failures (such as GitHub authentication issues and missing dependencies).
- Running `./runtests` causes some files to be rewritten; this is fine within a travis job but annoying when running locally if you have uncommitted changes.
 
- Adding `check_github` to `bin/pre-flight-checks` which is called by `make check`.
- Prevent `./runtests` from running if there are uncommitted changes

3. How can you test this?

- Run: `GITHUB_TOKEN=foo make check` (within an environment subdirectory)
- Run `./runtests` locally with and without uncommitted changes; the former should abort while the latter should proceed (`git checkout .` do reset local working tree)

- if it works for you (and if not, what goes wrong)
- if you think it's merge-worthy

### DIFF
--- a/bin/pre-flight-checks
+++ b/bin/pre-flight-checks
@@ -9,6 +9,26 @@ stderr_echo() {
   echo >&2 "$@"
 }
 
+check_github() {
+  echo "Attempting to authenticate with GitHub..."
+  : "${GITHUB_ORG:=travis-infrastructure}"
+  : "${GITHUB_TEAM:=2420497}"
+
+  auth="${GITHUB_USERNAME}:${GITHUB_TOKEN}"
+  cmd="curl --silent \
+    --output /dev/null \
+    --write-out '%{http_code}' \
+    https://${auth}@api.github.com/teams/${GITHUB_TEAM}/members"
+  status_code=$(eval "$cmd")
+  if test "$status_code" -eq 200; then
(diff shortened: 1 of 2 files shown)


runtests
```diff
@@ -4,6 +4,9 @@ set -o errexit
 set -o pipefail
 
 main() {
+  # abort if there are uncommitted local changes
+  git diff-index --quiet HEAD -- || (echo "You have uncommitted changes. Not proceeding, as this script changes some files in place." && exit 1)
```
meatballhat: I would like this wrapped around into more lines that are shorter :grin:

bin/pre-flight-checks
```diff
@@ -9,6 +9,26 @@ stderr_echo() {
...
 
+check_github() {
+  echo "Attempting to authenticate with GitHub..."
+  : "${GITHUB_ORG:=travis-infrastructure}"
+  : "${GITHUB_TEAM:=2420497}"
+
+  auth="${GITHUB_USERNAME}:${GITHUB_TOKEN}"
+  cmd="curl --silent \
+    --output /dev/null \
+    --write-out '%{http_code}' \
```
meatballhat: `--write-out`  TIL!

Issue #185, opened 2017-07-08
runtests not validating all bash scripts


edmorley:
I was looking for prior art for the testing of Terraform configs, which led to me borrowing parts of the runtests script in this repo for mozilla-platform-ops/devservices-aws/pull/55.

However whilst working on that I noticed that the current implementation only searches for scripts with the shebang `#!/usr/bin/env bash` - and whilst that's the best practice shebang, it's not unknown for people to use `#!/bin/bash`, `#!/usr/bin/bash` or variants with `sh` instead of `bash`.

And in fact, there are scripts in this repo that use these alternative shebangs:
https://github.com/travis-infrastructure/terraform-config/search?q="%2Fbin%2Fbash"

I'd recommend using this instead:
`git grep -El '^#!/.+\b(bash|sh)\b'`
(note the extended regex param)

Testing briefly against this repo using the new regex results in failures in the previously unchecked files - which I was going to open a PR to resolve, but some of the failures are due to variables set by Terraform that shellcheck can't see - so not sure how you'd want to handle that.

Note also that newer versions of shellcheck appears to find more issues in the rest of the repo too (initially I was testing locally with a newer version than that used in the current Travis run) - might be worth updating at some point.

edmorley:
Ah I forgot to paste an example failing job:
https://travis-ci.org/edmorley/terraform-config/builds/251412172

edmorley:
> Note also that newer versions of shellcheck appears to find more issues in the rest of the repo too  ... - might be worth updating at some point.

In case it saves anyone time in the future (since it confused me for a bit this morning for mozilla-platform-ops/devservices-aws#55)...

If switching from the custom built binaries in the travis-blue-public S3 bucket to the ones produced upstream (such as https://storage.googleapis.com/shellcheck/shellcheck-v0.4.6.linux.x86_64.tar.xz), make sure to add `--strip-components=1` to the `tar` invocation, otherwise the shellcheck binary ends up in a subdirectory and so not on PATH, which is easy to miss since then the older preinstalled shellcheck binary in /usr/bin/local is silently used instead.

meatballhat:
Resolved via #209

PR #209, opened 2017-08-22
shfmt all the things

soulshake:

Inconsistent formatting and (format-checking) of bash files

- upgrade `shfmt` in `.travis.yml`
- fix bash formatting to match more recent version of `shfmt`

3. How can you test this?

- you could run `./runtests` and ensure it does not result in any local modifications (make sure you have the most recent version of `shfmt`: `go get -u mvdan.cc/sh/cmd/shfmt`)

## Important

I fixed a number of quoting issues that were present in bash files that weren't previously being checked. However, there are a few I wasn't sure how to address. I worked around this by [disabling this check](https://github.com/travis-infrastructure/terraform-config/pull/209/files#diff-59d6da2a849342596840d50ffd6375e6R32):

`$ git grep -l '^#!/bin/bash' | xargs shellcheck -e SC2154`

You can run the above command locally without `-e SC2154` to view the warnings raised by shellcheck:

`$ git grep -l '^#!/bin/bash' | xargs shellcheck`

### DIFF
--- a/runtests
+++ b/runtests
@@ -28,9 +28,12 @@ main() {
 
   echo 'Running shellcheck'
   git grep -l '^#!/usr/bin/env bash' | xargs shellcheck
+  # Temporarily disable shellcheck checks for SC2154 (undefined variable) on /bin/bash files
+  git grep -l '^#!/bin/bash' | xargs shellcheck -e SC2154
 
   echo 'Running shfmt'
   git grep -l '^#!/usr/bin/env bash' | xargs shfmt -i 2 -w
+  git grep -l '^#!/bin/bash' | xargs shfmt -i 2 -w
 
   echo 'Running bats'
   for f in $(git ls-files '*.bats'); do
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ gce-*/config/
 graph.png
 local-development-0/local-development-0.tfstate
 macstadium-*/config/
+tfstate
+debug.tmp
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
   global:
   - PATH="${HOME}/bin:${PATH}"
   - SHELLCHECK_URL="https://s3.amazonaws.com/travis-blue-public/binaries/ubuntu/14.04/x86_64/shellcheck-0.4.4.tar.bz2"
-  - SHFMT_URL="https://github.com/mvdan/sh/releases/download/v0.1.0/shfmt_v0.1.0_linux_amd64"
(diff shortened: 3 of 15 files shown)


runtests
```diff
@@ -28,9 +28,12 @@ main() {
 
   echo 'Running shellcheck'
   git grep -l '^#!/usr/bin/env bash' | xargs shellcheck
+  # Temporarily disable shellcheck checks for SC2154 (undefined variable) on /bin/bash files
+  git grep -l '^#!/bin/bash' | xargs shellcheck -e SC2154
```
edmorley: This partly fixes #185 fwiw :-)

soulshake: Nice catch @edmorley, thanks!

Btw, I didn't include `/bin/sh` because `git grep` returns exit code 1 if it doesn't find any matches, so the script aborts because of `set -o errexit`. (We could pass `--no-run-if-empty` to `xargs` but it will still fail because of `set -o pipefail`).

Feel free to share a graceful solution if you're so inclined :)

mvdan:
A bit late, but note that you don't have to use `xargs` - you can simply `shfmt <dir>` and it will recursively act on all files that either have a shell extension or a shell shebang.

Also, this switches to using master. Which is fine - props to you for testing it :) - but it does mean CI might suddenly break, even if no breaking changes are done. For example, if a bug is fixed and the format is improved somewhere. Or, sometime in the future, if 3.0 comes along and a breaking change happens (like pipes and 2.0 in this diff).

soulshake:
Hi @mvdan, thanks for your tip and for providing this useful tool!

When I try `shfmt -i 2 -w .` I get a bunch of errors like:
```
open aws-precise-production-2/.terraform/modules/11f2ac1a7eb07b16bf06a65835575e56: is a directory
```
(Note: `aws-precise-production-2/.terraform/modules/11f2ac1a7eb07b16bf06a65835575e56` is a symlink to a directory.)

Let me know if this isn't expected behavior; happy to open an issue if needed.

Regarding the change to master (as opposed to a specific version of `shfmt`): I suppose we'll take our chances for now, and revert to getting a specific version if it causes problems. (Still, I'm in favor of the philosophy described [here](https://stackoverflow.com/questions/24855081/how-do-i-import-a-specific-version-of-a-package-using-go-get/26546559#26546559). :))

mvdan:
> Let me know if this isn't expected behavior; happy to open an issue if needed.

Ah, yes, sounds like a bug. I don't have tests with symlinks. Please do open an issue with details.

soulshake:
@mvdan Issue + PR opened: https://github.com/mvdan/sh/pull/153

PR #210, opened 2017-08-23
Identify shell files more classily

### LINKED ISSUE #185: runtests not validating all bash scripts

soulshake:

Resolves #185 

Use extended regex to check for shell files (as suggested by @edmorley)

3. How can you test this?

Run `./runtests` or `git grep -El '^#!/.+\b(bash|sh)\b'` in the repo root to view files identified as bash.

### DIFF
--- a/runtests
+++ b/runtests
@@ -31,13 +31,11 @@ main() {
   done
 
   echo 'Running shellcheck'
-  git grep -l '^#!/usr/bin/env bash' | xargs shellcheck
-  # Temporarily disable shellcheck checks for SC2154 (undefined variable) on /bin/bash files
-  git grep -l '^#!/bin/bash' | xargs shellcheck -e SC2154
+  # Note: shellcheck checks for SC2154 (undefined variable) are disabled by '-e SC2154' below
+  git grep -El '^#!/.+\b(bash|sh)\b' | xargs shellcheck -e SC2154
 
   echo 'Running shfmt'
-  git grep -l '^#!/usr/bin/env bash' | xargs shfmt -i 2 -w
-  git grep -l '^#!/bin/bash' | xargs shfmt -i 2 -w
+  git grep -El '^#!/.+\b(bash|sh)\b' | xargs shfmt -i 2 -w
 
   echo 'Running bats'
   for f in $(git ls-files '*.bats'); do


runtests
```diff
@@ -27,13 +27,11 @@ main() {
   done
 
   echo 'Running shellcheck'
-  git grep -l '^#!/usr/bin/env bash' | xargs shellcheck
-  # Temporarily disable shellcheck checks for SC2154 (undefined variable) on /bin/bash files
-  git grep -l '^#!/bin/bash' | xargs shellcheck -e SC2154
+  # Note: shellcheck checks for SC2154 (undefined variable) are disabled by '-e SC2154' below
+  git grep -El '^#!/.+\b(bash|sh)\b' | xargs shellcheck -e SC2154
```
meatballhat: I want to make sure we still check for undefined variables most of the time.  It's not documented here, but this was my original intent with using different shebangs (yes, a not great decision...).  I would prefer all files to manage individual shellcheck rules locally via comments.

soulshake: @meatballhat agreed; I did this as a workaround for the fact that some of the `.sh` files that weren't getting checked before raised `shellcheck` errors that I don't know how to fix.

edmorley: Perhaps instead individual SC2154 occurrences could be ignored using the per-file annotations?
https://github.com/koalaman/shellcheck/wiki/Ignore#ignoring-one-specific-instance-in-a-file
https://github.com/koalaman/shellcheck/wiki/Ignore#ignoring-all-instances-in-a-file-044

soulshake: Oh nice!

PR #219, opened 2017-09-08
Remove aws-precise-staging-1 infrastructure

meatballhat:
- [x] `make destroy && make apply`


Lyoness:
`MAKE DESTROY` SHOULD BE IN ALL CAPS AS IT IS THE MOST METAL COMMAND EVUR :smiling_imp:

PR #222, opened 2017-09-14
Dynamic step scaling

soulshake:
I've been testing @igorwwwwwwwwwwwwwwwwwwww's step scaling setup on staging.

I found that having multiple steps helped me reason about the thresholds more easily better, so I initially hardcoded a few and it seems to work nicely on staging.

But when applied to both staging and production, it makes more sense to scale relative to headroom, so I switched to using fractions (which probably makes it less readable to most of us).

Note: When I applied @igorwwwwwwwwwwwwwwwwwwww's original PR to staging and tried executing the `staging-1-workers-org-add-capacity` policy manually, I got a 400 error in the AWS console (I didn't note it, unfortunately, but I think it was about a missing upper bound or something like that).


soulshake:
Seems to be working as expected in staging:

step-scaling-launching-while-terminating

modules/aws_asg/main.tf
```diff
@@ -332,9 +332,23 @@ resource "aws_autoscaling_policy" "workers_add_capacity" {
   estimated_instance_warmup = "${var.worker_asg_scale_out_cooldown}"
   metric_aggregation_type   = "Maximum"
 
+  # Headroom is just below THRESHOLD, scale out normally
   step_adjustment {
     scaling_adjustment          = "${var.worker_asg_scale_out_qty}"
-    metric_interval_lower_bound = 1.0
+    metric_interval_lower_bound = "${var.worker_asg_scale_out_threshold * -1/3 -1}"
```
meatballhat: I don't understand the intent here 😕 

When `worker_asg_scale_out_threshold` has a value of `80`, as is the case for one of the ASGs in production, then `${var.worker_asg_scale_out_threshold * -1/3 -1}` gives me `-27.6`.  Does terraform get a different result?  What _should_ it be?

soulshake: Something like "when we're below the threshold, create n instances. When we're at 50% of the threshold, create n * 2 instances."
I pushed another commit with less weird torturous math.

modules/aws_asg/main.tf
```diff
@@ -332,9 +332,16 @@ resource "aws_autoscaling_policy" "workers_add_capacity" {
...
   step_adjustment {
     scaling_adjustment          = "${var.worker_asg_scale_out_qty}"
-    metric_interval_lower_bound = 1.0
+    metric_interval_lower_bound = "${floor(var.worker_asg_scale_out_threshold/-2.0)}"
+  }
+
+  # Headroom is less than half of THRESHOLD; scale out twice as much
+  step_adjustment {
+    scaling_adjustment          = "${var.worker_asg_scale_out_qty * 2}"
+    metric_interval_upper_bound = "${floor(var.worker_asg_scale_out_threshold/-2.0)}"
```
meatballhat: Do the negative interval bounds translate as drops in headroom?  What do the `metric_interval_lower_bound` and `metric_interval_upper_bound` usually look like?

soulshake: The bounds are relative to the threshold. So if the threshold is 10, and the lower bound of one action is -4, then that action will execute when the level is above 6.
If there's an action with an *upper* bound of -4, then it will execute when the level drops _below_ 6.

soulshake: For the threshold `v1.travis.rabbitmq.consumers.staging.builds.ec2.headroom < 8 for 2 minutes`

This is what it translates to: 

step-scaling

If the bounds were positive, it would instead look like this:

step-scaling

meatballhat: I think I understand better ⁉️ 😖

soulshake:
This illustrates the ability to launch new instances even while some are terminating (unlike yesterday, when I tested by manually changing the desired size before launching a bunch of jobs, this was purely job-triggered on both the scale-in and scale-out):

step-scaling-working

soulshake:
## Before
old-scaling-policies-org

## After
new-scaling-policies-org

PR #223, opened 2017-09-14
Increase TRAVIS_WORKER_POOL_SIZE from 13 to 17

soulshake:

Since fixing [worker CPU allocation](https://github.com/travis-ci/worker/pull/366), workers are typically using less than 50% of their CPU:

unutilized-cpu

The PR increases the number of jobs that can run per instance (`TRAVIS_WORKER_POOL_SIZE`) from 13 to 17.

The worker instance type is `c3.8xlarge` which has 32 vCPUs. We should therefore be able to run 16 jobs (since the worker container CPUset is not limited). I'd like to bump to 17 (1 more than should be possible due to CPU constraints) since worker should reject jobs if there aren't enough free CPU sets anyway.

Is everything I said above accurate?

Many jobs don't take advantage of having multiple CPUs available to them. Any ideas on an approach that would allow better utilization of resources without letting some jobs slow others down?


meatballhat:
🤔

PR #226, opened 2017-09-15
Separate out secret/nonsecret bits for gce-production-1

meatballhat:

All of the configuration is kept in a secret place, and duplicated for org and com.  I much prefer the way stuff is implemented in the AWS infrastructures.  This is a step.

Separate out the nonsecret bits into a locally-tracked file as with the AWS infras.

## How can you test this?

Terraform plan + apply :boom:

Anybody against this?

- [ ] secret PR for com is ready
- [ ] secret PR for org is ready


meatballhat:
not-boom 🎉 

```
Apply complete! Resources: 8 added, 24 changed, 4 destroyed.
```

PR #235, opened 2017-09-19
Add use of config-writing script to GCE envs

meatballhat:
plus some other little bits I noticed along the way.

The GCE and AWS infrastructures remain slightly inconsistent in how secret and non-secret configuration bits are ingested.

I decided to use the same config writing script that the AWS infrastructures use, ported to rubby for arguably better UX and maintainability, since it is known to work.

## How can you test this?

`make plan` in any GCE or AWS project to ensure stuff isn't busted.


erickinganco:
Wow, I don't understand any of this, does anyone speak English on here, would really like to find people interested in my ideas......

erickinganco:
Fyi I'm jay btw....

PR #238, opened 2017-09-19
Run 3x workers on GCE instances

meatballhat:

We currently run 1x `travis-worker` process per GCE instance, each being a `g1.small` machine type.  These machines are barely utilized, and the recent shift to run a smaller pool size per worker to reduce the impact of worker restarts means that we're utilizing them even less.

Create 3x `travis-worker` upstart (or systemd) configs, making 3x the worker processes per instance.

## How can you test this?

GCE staging fun times.

Should the multiplier be configurable, or is 3x per instance a reasonable start?  Other?

### DIFF
--- a/assets/travis-worker/travis-worker-wrapper
+++ b/assets/travis-worker/travis-worker-wrapper
@@ -2,12 +2,14 @@
 set -o errexit
 
 main() {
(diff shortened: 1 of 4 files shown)


meatballhat:
👍 This is working in gce-staging-1

PR #243, opened 2017-09-20
Add commit hash to Slack notification

soulshake:

The Slack notification when one runs `make apply` should include the git commit.

Add the commit to the output, e.g. `branch @ commit`.

## How can you test this?

Run it!

### DIFF
--- a/bin/post-flight
+++ b/bin/post-flight
@@ -44,7 +44,7 @@ EOF
 }
 
 __git_branch_at_commit() {
-  printf "%s @ %s" "$(git rev-parse --abbrev-ref HEAD)" "$(git log --oneline -1)"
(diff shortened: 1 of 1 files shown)


bin/post-flight
```diff
@@ -43,8 +43,8 @@ EOF
   fi
 }
 
-__git_branch() {
-  git rev-parse --abbrev-ref HEAD
+__git_branch_at_commit() {
+  printf "%s @ %s" "$(git rev-parse --abbrev-ref HEAD)" "$(git log --oneline -1)"
```
meatballhat: This makes for a pretty long line, at least for me.  Can we instead do `git log --format='%h' -1`?

PR #245, opened 2017-09-20
Pre-seed images on worker instances via EBS volume

soulshake:

10GB+ of images being pulled at runtime slows down the bootstrapping process significantly.
 
Attaching an EBS volume created from a snapshot containing the images so the `docker pull` at runtime only fetches new bits.

## How can you test this?

`make apply` in staging and check a new instance.


meatballhat:
> Attaching an EBS volume created from a snapshot containing the images so the docker pull at runtime only fetches new bits.

@soulshake Have you observed this?  My assumption has been that the delta between the snapshot docker images and different docker image tags in the worker prestart script will result in a fairly large pull, since the only layers shared between these (at best) are the underlying ubuntu base.  This ubuntu base image is a small portion of the image size, coming in at ~50MB out of ~3GB.  Any time the upstream ubuntu base changes, this slim chance of a shared image layer is cache-busted.

soulshake:
@meatballhat Actually... Would that still be an issue if we generate an updated snapshot anytime there's a new docker image tag?

The scenario I'm envisioning:
* New snapshots could be generated when a new docker image tag is detected (How? TBD but I assume we can automate this)
* New instances would be bootstrapped with the updated snapshot
* Existing instances could be made to pull the new layers without interrupting running jobs (if we want)

meatballhat:
Container-based Linux is so 2014 💔

PR #250, opened 2017-10-03
Use GCE resources differently/better

meatballhat:

We are receiving `ZONE_RESOURCE_POOL_EXHAUSTED` errors for us-central1-c and `IP_SPACE_EXHAUSTED` for the `workerscom` subnetwork.

Use all 4 available zones in the us-central1 region so that we can reduce the likelihood of zone resource pool errors.  Create and assign a larger subnetwork for com (and org) jobs.

## How can you test this?

- [x] `make plan`
- [x] `make apply`

Downsides?


assets/travis-worker/travis-worker-wrapper
```diff
@@ -30,4 +39,11 @@ main() {
     "${TRAVIS_WORKER_SELF_IMAGE}" travis-worker
 }
 
+__rand_gce_zone() {
+  local zones idx
+  zones=( ${1//,/ } )
```
bogdanap: The spaces here seem to make Travis sad.

meatballhat:
@bogdanap regarding instance count adjustments:

- ensure each count is a multiple of 4, given we are using 4 zones now instead of 2
- increase com worker capacity overall
- shift usage to "production-1" (eco-emissary-99515) whenever possible

PR #251, opened 2017-10-06
Check docker health with a cron job

soulshake:

`docker` sometimes gets into a weird state on worker instances.

This PR adds an hourly cron job to check that docker is still responsive. If not, the instance is imploded.

## How can you test this?

We'll see on staging?

* Is there a better way?
* Should be consider rebooting instead of shutting down?

Addresses https://github.com/travis-pro/team-blue/issues/704

### DIFF
--- a/modules/aws_asg/check-docker-health.bash
+++ b/modules/aws_asg/check-docker-health.bash
@@ -59,7 +59,7 @@ __handle_unresponsive_docker() {
   sleep "${post_sleep}"
 
   if [ -e "${run_d}/implode" ]; then
-    "${KILL_COMMAND}" -INT "$(pidof travis-worker)" || restart travis-worker
+    "${KILL_COMMAND}" -TERM "$(pidof travis-worker)" || restart travis-worker
   else
     logger "${run_d}/implode not found, not imploding?"
     __die noop 0
--- a/modules/aws_asg/check-docker-health.bash
+++ b/modules/aws_asg/check-docker-health.bash
@@ -34,6 +34,12 @@ main() {
     __handle_unresponsive_docker "${run_d}"
     __die imploding 86
   fi
+
+  if [ -e "${run_d}/implode" ]; then
+    logger "docker no longer seems unhealthy; canceling implosion."
+    rm "${run_d}/implode"
+  fi
+
   __die noop 0
 }
 
--- a/modules/aws_asg/check-docker-health.bash
+++ b/modules/aws_asg/check-docker-health.bash
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o pipefail
+
+# Sometimes, the docker service will be running, but certain commands (docker ps) will hang indefinitely.
+# This script detects this behavior and implodes the instance when it occurs.
+
+main() {
+  local warmup_grace_period=600
+  local post_sleep="${POST_SHUTDOWN_SLEEP}"
+  local sleep_time="${DOCKER_PS_SLEEP_TIME}"
+  local run_d="${RUNDIR}"
+  : "${post_sleep:=300}"
+  : "${sleep_time:=5}"
+  : "${run_d:=/var/tmp/travis-run.d}"
+  : "${KILL_COMMAND:=kill}"
+
+  if [[ -f "${run_d}/implode.confirm" ]]; then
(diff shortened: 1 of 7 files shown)


modules/aws_asg/check-docker-health.bash
```diff
@@ -0,0 +1,65 @@
...
+}
+
+__handle_implode_confirm() {
+  local run_d="${1}"
+  local post_sleep="${2}"
+
+  local reason
+  reason="$(cat "${run_d}/implode.confirm" 2>/dev/null)"
+  : "${reason:=not sure why}"
+  echo "${SHUTDOWN:-/sbin/shutdown}" -P now "imploding because ${reason}"
```
meatballhat: Is the leading `echo` intentional?

modules/aws_asg/check-docker-health.bash
```diff
@@ -0,0 +1,65 @@
...
+  sleep "${post_sleep}"
+}
+
+__handle_unresponsive_docker() {
+  local run_d="${1}"
+  echo "docker appears to be unhealthy" |
+    tee "${run_d}/implode"
+  "${DOCKER:-docker}" kill -s INT travis-worker
```
meatballhat: If docker is unresponsive, won't this `docker kill` block?  If the instance is being imploded, I think this line can be safely removed 😺

runtests
```diff
@@ -5,8 +5,8 @@ set -o pipefail
 
 main() {
   # abort if there are uncommitted local changes
-  msg="You have uncommitted changes. Not proceeding, as this script changes some files in place."
-  git diff-index --quiet HEAD -- || (echo "$msg" && exit 1)
+  #msg="You have uncommitted changes. Not proceeding, as this script changes some files in place."
+  #git diff-index --quiet HEAD -- || (echo "$msg" && exit 1)
```
meatballhat: I wonder if we should add a check to bypass the changes check so that `./runtests` may be used more easily while working on a change 🤔

modules/aws_asg/check-docker-health.bash
soulshake: Touche, removed

modules/aws_asg/check-docker-health.bash
soulshake: Whoops, nice catch

soulshake:
[This test](https://travis-ci.org/travis-infrastructure/terraform-config/builds/292492180#L434) is failing. I think it might be because of trying to mock a builtin (`kill`). If so, I'm not sure how to work around it.

runtests
```diff
@@ -5,8 +5,10 @@ set -o pipefail
 
 main() {
   # abort if there are uncommitted local changes
-  msg="You have uncommitted changes. Not proceeding, as this script changes some files in place."
-  git diff-index --quiet HEAD -- || (echo "$msg" && exit 1)
+  if [[ "$1" != "--force" ]]; then
```
meatballhat: Instead of checking `$1`, which is also used below for `--env <env-file>`, what about `if [[ "${*}" =~ --force ]]; then` to see if `--force` is anywhere in args?

modules/aws_asg/check-docker-health.bats
```diff
@@ -0,0 +1,113 @@
...
+  bash "${BATS_TEST_DIRNAME}/check-docker-health.bash"
+}
+
+@test "handles implosion confirmation" {
+  cat >"${BATS_TMPDIR}/returns/date" <<EOF
+20171030T153252
+EOF
+  echo 'docker appears to be unhealthy' >"${RUNDIR}/implode.confirm"
+  run run_check_docker_health
+  KILL_COMMAND="${BATS_TMPDIR}/bin/kill_mocked"
```
meatballhat: Why is `KILL_COMMAND` defined here?

modules/aws_asg/check-docker-health.bats
```diff
@@ -0,0 +1,113 @@
...
+  cat >"${RUNDIR}/implode" <<EOF
+docker appears to be unhealthy
+EOF
+
+  cat >"${BATS_TMPDIR}/returns/date" <<EOF
+20171030T153252
+EOF
+
+  run run_check_docker_health
+  cat /tmp/logs/mock.log
```
meatballhat: Is this `cat /tmp/logs/mock.log` a leftover from debugging?

PR #255, opened 2017-10-13
define site tag for gcp

igorwwwwwwwwwwwwwwwwwwww:

This is the necessary worker config for travis-ci/worker#389.


meatballhat:
We already define + internally recognize "Travis site", which have the same values as you've defined here.  Can you use the existing value instead?

igorwwwwwwwwwwwwwwwwwwww:
Closing, as the worker PR is being reworked to re-use the global config. 👍

PR #258, opened 2017-10-20
Migrate capacity from gce-production-{3,4,5,6} to gce-production-1

meatballhat:
now that we have significantly higher resource quotas in
gce-production-1.

The existence of 6 different production GCE projects, contributing to:
- administrative & cognitive overhead
- likelihood of human error
- 😿 

Before writing this e-note: requesting an increase in API quota, which was granted in July, and more recently requesting a sizeable increase in resource quotas, which was granted on Thursday, October 19th.

Shift both org and com capacity from gce-production-{3,4,5,6} to gce-production-1, additionally increasing the per-instance job capacity from 45-54 to 90.

## How can you test this?

`make plan`, `make apply` 🎉


igorwwwwwwwwwwwwwwwwwwww:
Huge 👍 

One note: I believe the subnets need to be increased to prevent ip address space exhaustion.

meatballhat:
@igorwwwwwwwwwwwwwwwwwwww Already addressed 🎉  https://github.com/travis-infrastructure/terraform-config/commit/31b024b436e2e4f23892c681385dae4b1c309187

PR #277, opened 2017-11-04
Enable docker debug mode

soulshake:

We have no helpful logs on instances where docker has frozen, and by the time docker freezes, it's too late to enable debug mode.

This PR will enable debug mode for docker on all worker instances.

**Important:** Debug mode is extremely noisy due to how often travis-worker interacts with containers. We should keep an eye on disk space and other side effects (Papertrail?), and revert this soon.

### DIFF
--- a/modules/aws_asg/main.tf
+++ b/modules/aws_asg/main.tf
@@ -152,7 +152,7 @@ EOF
 data "template_file" "docker_daemon_json" {
   template = <<EOF
 {
-  "graph": "/mnt/docker",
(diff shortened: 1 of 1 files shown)


modules/aws_asg/main.tf
```diff
@@ -152,7 +152,7 @@ EOF
 data "template_file" "docker_daemon_json" {
   template = <<EOF
 {
-  "graph": "/mnt/docker",
+  "data-root": "/mnt/docker",
```
meatballhat: 👏

PR #278, opened 2017-11-04
Empower check-docker-health.bash to confirm implosion

soulshake:

Normally, `/var/tmp/travis-run.d/implode.confirm` is created by travis-worker when it finds `/var/tmp/travis-run.d/implode` upon starting.

If docker is stuck, then travis-worker doesn't necessarily have a PID, so we can't send it a `TERM` signal, so it can't start, meaning `implode.confirm` will never be created.

Allow `check-docker-health.bash` to create `implode.confirm` if `travis-worker` doesn't have a PID.

## How can you test this?

This can't really be tested on staging since we can't reproduce [the docker issue](https://github.com/moby/moby/issues/35408) this is intended to address.

### DIFF
--- a/modules/aws_asg/check-docker-health.bash
+++ b/modules/aws_asg/check-docker-health.bash
@@ -7,16 +7,16 @@ set -o pipefail
 
 main() {
   local warmup_grace_period=600
-  local post_sleep="${POST_SHUTDOWN_SLEEP}"
+  local pre_implosion_sleep="${POST_SHUTDOWN_SLEEP}"
   local sleep_time="${DOCKER_PS_SLEEP_TIME}"
   local run_d="${RUNDIR}"
-  : "${post_sleep:=300}"
+  : "${pre_implosion_sleep:=300}"
   : "${sleep_time:=5}"
   : "${run_d:=/var/tmp/travis-run.d}"
   : "${KILL_COMMAND:=kill}"
 
   if [[ -f "${run_d}/implode.confirm" ]]; then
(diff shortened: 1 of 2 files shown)


modules/aws_asg/check-docker-health.bash
```diff
@@ -1,22 +1,21 @@
 #!/usr/bin/env bash
-set -o errexit
```
meatballhat: Why does this need to be removed?

modules/aws_asg/check-docker-health.bash
```diff
@@ -45,32 +44,37 @@ main() {
...
   local run_d="${1}"
-  local post_sleep="${2}"
+  local pre_implosion_sleep="${2}"
 
   local reason
   reason="$(cat "${run_d}/implode.confirm" 2>/dev/null)"
   : "${reason:=not sure why}"
   "${SHUTDOWN:-/sbin/shutdown}" -P now "imploding because ${reason}"
-  sleep "${post_sleep}"
 }
```
meatballhat: I see the `pre_implosion_sleep` local var declared here, but used (undeclared) in `__handle_unresponsive_docker` 😕 

IIRC the point of the previous `sleep "${post_sleep}"` was to allow time for `shutdown` to take effect without re-running all these checks immediately, which made sense in the context of `travis-worker` hook scripts, but I can see how it's not necessary here 👍

soulshake: Good catches :/

PR #280, opened 2017-11-07
Cha remove custom3

Lyoness:

Schibsted is no longer using their dedicated macpros, aliased to custom-3

See https://travisci.slack.com/archives/C45D4L2QP/p1509720277000021
and https://travisci.slack.com/archives/C45D4L2QP/p1509714699000430
for reference.

Also see: https://github.com/travis-pro/travis-keychain/pull/378

,
Since we have no replacement for custom-3, delete it.  Keep custom-{4,5,6} same. 

## How can you test this?

`make plan` locally and see if it blends. (provided we can get a working version of Terraform :confounded: )

Any feedback is welcome. :+1:

### DIFF
--- a/macstadium-shared-1/Makefile
+++ b/macstadium-shared-1/Makefile
@@ -8,7 +8,6 @@ CONFIG_FILES := \
 		config/jupiter-brain-staging-org-env \
 		config/jupiter-brain-custom-1-env \
 		config/jupiter-brain-custom-2-env \
-		config/jupiter-brain-custom-3-env \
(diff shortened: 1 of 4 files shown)


Lyoness:
Excellent point @emdantrim 

This change was applied in the `macstadium-shared-1` environment with a `make-destroy-apply` to wipe out previous remote state, and `make-apply` to rebuild anew. 

However the changes were _not_ picked up in `macstadium-shared-2`

PR #282, opened 2017-11-07
Initial commit for ubuntu 16.04

soulshake:

**Work in progress** Upgrade to Ubuntu 16.04, to try to address https://github.com/moby/moby/issues/35408

14.04 uses upstart, whereas 16.04 uses systemd, so some rearrangement is needed.

## How can you test this?

Not testable yet.

Should we maintain compatibility with upstart for a while, or just switch completely over?

Should we maintain direct-lvm, or switch to overlay2, or make it switchable between them?


meatballhat:
I'm pretty sure this PR is no longer relevant 🙀

soulshake:
Indeed!

PR #285, opened 2017-11-09
Increase TRAVIS_WORKER_DOCKER_INSPECT_INTERVAL from 500ms to 1000ms

soulshake:

Frequent exec inspects _might_ be aggravating [this docker issue](https://github.com/moby/moby/issues/35408).

Double the interval between exec inspections.

## How can you test this?

Apply in staging, check `docker.log`.


meatballhat:
@soulshake FYI, because this value is parsed as a duration, it can be `1s` instead of `1000ms` 😁

PR #288, opened 2017-11-11
Add a cron job to prune old containers

soulshake:

For some reason, some containers continue running well beyond the maximum time limit for jobs.

This PR adds a cron job to check for, and remove, any containers that are older than three hours (10800 seconds).

## How can you test this?

I don't know how to reproduce the issue of containers living beyond their expected time. But one could apply in staging and monitor the logs for killed containers.

Is it safe to assume that containers killed in this way will not get requeued?

### DIFF
--- a/modules/aws_asg/kill-old-containers-crontab
+++ b/modules/aws_asg/kill-old-containers-crontab
@@ -1,3 +1,3 @@
 SHELL = /bin/bash
 
-* * * * * root /var/tmp/travis-run.d/kill-old-containers >> /var/log/syslog
+* * * * * root /var/tmp/travis-run.d/kill-old-containers &>> /var/log/syslog
--- a/modules/aws_asg/kill-old-containers-crontab
+++ b/modules/aws_asg/kill-old-containers-crontab
@@ -1,3 +1,3 @@
 SHELL = /bin/bash
 
-* * * * * root /var/tmp/travis-run.d/kill-old-containers &>> /var/log/syslog
+* * * * * root /var/tmp/travis-run.d/kill-old-containers >> /var/log/syslog
--- a/modules/aws_asg/kill-old-containers-crontab
+++ b/modules/aws_asg/kill-old-containers-crontab
@@ -1,0 +1,3 @@
+SHELL = /bin/bash
+
+* * * * * root /var/tmp/travis-run.d/kill-old-containers >> /var/log/syslog
--- a/modules/aws_asg/kill-old-containers.bash
+++ b/modules/aws_asg/kill-old-containers.bash
@@ -20,7 +20,8 @@ __report_kills() {
 
   # request_body=$(< <(cat <<EOF)
   # read -r -d '' request_body <<EOF
-  request_body=$(cat <<EOF
+  request_body=$(
+    cat <<EOF
   { "measure_time": "$timestamp",
     "source": "cron.ec2.aj.container-killer",
     "gauges": [
--- a/modules/aws_asg/kill-old-containers.bash
+++ b/modules/aws_asg/kill-old-containers.bash
@@ -17,21 +17,27 @@ __report_kills() {
   count_killed="$1"
   count_not_killed="$2"
   timestamp="$(date +%s)"
+  site="$TRAVIS_WORKER_TRAVIS_SITE"
+  stage="staging"
+  if [[ "$HOSTNAME" == *"production"* ]]; then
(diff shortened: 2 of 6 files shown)


modules/aws_asg/kill-old-containers.bash
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -e
```
meatballhat: Given the subshells, how about `set -o pipefail`, too?

modules/aws_asg/kill-old-containers-crontab
```diff
@@ -0,0 +1,3 @@
+SHELL = /bin/bash
+
+* * * * * root /var/tmp/travis-run.d/kill-old-containers >> /var/log/syslog
```
meatballhat: nitpick: `&>>` to send stderr, too?

modules/aws_asg/kill-old-containers.bash
```diff
@@ -0,0 +1,29 @@
...
+  local code="${2}"
+  local count="${3}"
+  logger "time=$(date -u +%Y%m%dT%H%M%S) " \
+    "prog=$(basename "${0}") status=${status} count=${count}"
+  exit "${code}"
+}
+
+main() {
+  worker_cid=$(docker inspect travis-worker --format '{{ .Id }}')
+  cids=$(docker ps -q --filter before="$worker_cid")
```
meatballhat: nitpick: `worker_cid` and `cids` should both be declared as locals prior to subshell definition

modules/aws_asg/kill-old-containers.bash
```diff
@@ -0,0 +1,29 @@
...
+  local code="${2}"
+  local count="${3}"
+  logger "time=$(date -u +%Y%m%dT%H%M%S) " \
+    "prog=$(basename "${0}") status=${status} count=${count}"
+  exit "${code}"
+}
+
+main() {
+  local worker_cid=$(docker inspect travis-worker --format '{{ .Id }}')
+  local cids=$(docker ps -q --filter before="$worker_cid")
```
meatballhat: Sorry, I didn't explain this well 😅   Shellcheck gets upset about locals defined in-line via subshell due to issues with exit code handling, so splitting up the declarations and assignments is needed, a la:

``` bash
local somefoo
somefoo="$(docker inspect foo --format '{{ .Id }}')"
```

That's what these errors are about, at least: https://travis-ci.org/travis-infrastructure/terraform-config/builds/308032321#L434-L441

modules/aws_asg/kill-old-containers.bash
```diff
@@ -0,0 +1,116 @@
...
+  if [[ "$HOSTNAME" == *"production"* ]]; then
+    stage="production"
+  fi
+
+  # request_body=$(< <(cat <<EOF)
+  # read -r -d '' request_body <<EOF
+  request_body=$(
+    cat <<EOF
+  { "measure_time": "$timestamp",
+    "source": "cron.ec2.$site.$stage.aj-container-killer",
```
meatballhat: Do you want the metric to have your initials? 😁

modules/aws_asg/kill-old-containers.bash
```diff
@@ -0,0 +1,116 @@
+#!/bin/bash
+#set -e
```
meatballhat: No `set -e` / `set -o errexit`?

modules/aws_asg/kill-old-containers.bash
```diff
@@ -0,0 +1,116 @@
...
+  : "${LIBRATO_USERNAME:=${TRAVIS_WORKER_LIBRATO_EMAIL}}"
+  : "${LIBRATO_TOKEN:=${TRAVIS_WORKER_LIBRATO_TOKEN}}"
+  : "${LIBRATO_CREDENTIALS:=${LIBRATO_USERNAME}:${LIBRATO_TOKEN}}"
+  : "${RUNDIR:=/var/tmp/travis-run.d}"
+
+  local instance_id cids killed_count not_killed_count max_age
+  max_age="${MAX_AGE}"
+  killed_count=0
+  not_killed_count=0
+  cids=$(docker ps -q)
```
meatballhat: I would wrap this subshell in double quotes, if only for clarity about expecting space-separated values.

soulshake:
Thanks for feedback @meatballhat; I'll address these in another PR ASAP.

PR #293, opened 2017-11-14
Start haproxy service on wjb-* hosts

Lyoness:

Haproxy on `wjb-*` hosts doesn't automatically start after rebuilding hosts using terraform. 

Added a quick `sudo start` in the remote exec inline as a last step 

## How can you test this?
Since we don't have an isolated staging wjb host, I cannot. :disappointed:   but I want to merge it anyway for next time, so it is not a thing we forget. 

Is this the right approach?

### DIFF
--- a/modules/haproxy/haproxy.tf
+++ b/modules/haproxy/haproxy.tf
@@ -73,7 +73,7 @@ EOF
       "DEBIAN_FRONTEND=noninteractive sudo apt-get -y install haproxy",
       "sudo mv /tmp/haproxy.cfg /etc/haproxy/haproxy.cfg",
       "echo ENABLED=1 | sudo tee /etc/default/haproxy",
-      "sudo /etc/init.d/haproxy start",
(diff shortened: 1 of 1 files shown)


Lyoness:
https://github.com/travis-pro/team-blue/issues/770

modules/haproxy/haproxy.tf
```diff
@@ -73,6 +73,7 @@ EOF
       "DEBIAN_FRONTEND=noninteractive sudo apt-get -y install haproxy",
       "sudo mv /tmp/haproxy.cfg /etc/haproxy/haproxy.cfg",
       "echo ENABLED=1 | sudo tee /etc/default/haproxy",
+      "sudo /etc/init.d/haproxy start",
```
meatballhat: I prefer `sudo service haproxy start` 😁

Lyoness: done!

PR #305, opened 2017-11-29
Automate record/publish of NAT public IP addresses

meatballhat:

Published NAT IP addresses are currently manually updated, which tends to mean that they lag behind whenever changes are made, which results in sadness for anyone that depends on an accurate safelist of IP addresses for accessing resources across the public internet.

Generate and commit a JSON file similarly to the way we record language mappings, since the latter is already being used to automate docs.

## How can you test this?

It's done!

**Update**: please note that the IP addresses in question are Elastic IPs (EIP), but that the pool of EIPs that we "own" in our AWS account is a superset of those that we assign to NATs.  AFAICT, the way terraform works in the current setup results in these EIPs potentially around any time there are changes to the underlying NAT instances (which is rare).

### DIFF
--- a/aws-shared-2/generated-public-ip-addresses.json
+++ b/aws-shared-2/generated-public-ip-addresses.json
@@ -1,0 +1,36 @@
+{
+  "ips_by_host": [
+    {
+      "host": "workers-nat-com-shared-2.aws-us-east-1.travisci.net",
+      "ips": [
+        "34.234.4.53",
+        "52.45.220.64",
+        "54.208.31.17",
+        "52.54.40.118"
+      ]
+    },
+    {
+      "host": "workers-nat-org-shared-2.aws-us-east-1.travisci.net",
+      "ips": [
+        "52.3.55.28",
+        "34.233.56.198",
+        "52.54.31.11",
+        "52.45.185.117"
+      ]
+    }
+  ],
+  "hosts": [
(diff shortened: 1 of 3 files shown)


bogdanap:
Rather than doing this, could we look into using Elastic IPs, as @soulshake mentioned to prevent the public IPs from changing? I think it would save a lot of effort on the users' side.

meatballhat:
@bogdanap These are Elastic IPs, but the pool of Elastic IPs shifts around a bit given the current terraform setup.  I'm not sure how feasible it is to configure terraform to ensure a specific set of the available Elastic IPs are always used for specific hosts.

soulshake:
If the pool of EIPs that we "own" in our AWS account is a superset of those that we assign to NATs, could we just add our full list of owned EIPs to the docs? Or set up a hostname that resolves to either all of them, or the currently active ones, so users wouldn't need to manually update any time our NAT situation changes?

meatballhat:
@soulshake I am in favor of publishing all of our EIPs, which we could generate from an AWS query.  IIRC some customers depend on knowing the hostnames of the NATs, too.

@bogdanap Here are some of the tasks that integration with external data sources in the docs, for example: https://github.com/travis-ci/docs-travis-ci-com/blob/340a7151a14c939a77b5fdd9d90d1a95a20957c1/Rakefile#L84-L120

PR #309, opened 2017-12-06
New Macstadium Pods for the New Datacenter [WIP]

Lyoness:
WIP 

See https://github.com/travis-pro/team-blue/issues/736
We don't want to `terraform-apply` the existing `macstadium-shared-*` environments to the new datacenter.  
`cp -r` the `macstadium-shared-*` directories to `macstadium-pod-*` directories 

## How can you test this?
`make plan` , see if it works,  

- We have 1 additional SAN in each pod, so wondering if we need to declare a 2nd datastore variable for each env? 
- Document in README some of the workflow gotchas that come with having to use a 0.9.* release of terraform to get the vsphere provider that has been working for us. (Relevant to macstadium-* only envs)


Lyoness:
@ryndaniels and I were able to do a successful `make plan` on both `macstadium-pod-{1|2}` with identical outputs.  We'd like to merge this without doing a `make apply` as we prefer to add more work to the hosts before applying, while keeping PRs separate and small/thematic.

PR #315, opened 2017-12-15
Do not do double work

joepvd:
The recipe for the `$(CONFIG_FILES)` expands into a lot of targets. The recipe has been running for each of these targets on each invocation of `make plan`. This change runs the recipe once.

Ensure there is a single target for recipe that needs to be executed once.

## How can you test this?
Count how often lines from the recipes get into the output. `mkdir -p config`, for example.


meatballhat:
:scream_cat: C O N F L C I T :interrobang:

PR #318, opened 2017-12-20
[WIP] Increase pre-boot-sleep and limit upload retries

bogdanap:


bogdanap:
This caused trusty builds to start failing due to SSH errors and be requeued so I reverted the changes.

PR #326, opened 2018-01-22
Create DHCP server to serve the jobs subnet

emdantrim:
We want to use a DHCP server that is not pfSense to serve the jobs subnet.

Creating a DHCP server using Terraform.

## How can you test this?
Apply it! (Or check out how I've already applied it)

Are all the variables for the DHCP configuration acceptable?

### DIFF
--- a/modules/macstadium_infrastructure/dhcp-server.tf
+++ b/modules/macstadium_infrastructure/dhcp-server.tf
@@ -1,0 +1,50 @@
+resource "vsphere_virtual_machine" "dhcp_server" {
+  name       = "dhcp-server-${var.index}"
+  folder     = "${vsphere_folder.internal_vms.path}"
+  vcpu       = 2
+  memory     = 4096
+  datacenter = "${var.datacenter}"
+  cluster    = "${var.cluster}"
+  domain     = "macstadium-us-se-1.travisci.net"
+
+  network_interface {
+    label = "${var.internal_network_label}"
+  }
+
+  network_interface {
+    label              = "${var.jobs_network_label}"
+    ipv4_address       = "${cidrhost(var.jobs_network_subnet, 10 + var.index)}"
+    ipv4_prefix_length = "18"
+  }
+
+  disk {
+    template  = "${vsphere_folder.vanilla_vms.path}/${var.vanilla_image}"
+    datastore = "${var.datastore}"
+  }
+
+  connection {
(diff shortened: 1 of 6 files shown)


modules/macstadium_infrastructure/dhcp-server.tf
```diff
@@ -0,0 +1,50 @@
...
+    ipv4_address       = "${cidrhost(var.jobs_network_subnet, 10 + var.index)}"
+    ipv4_prefix_length = "18"
+  }
+
+  disk {
+    template  = "${vsphere_folder.vanilla_vms.path}/${var.vanilla_image}"
+    datastore = "${var.datastore}"
+  }
+
+  connection {
```
meatballhat: Is this `connection` used?

modules/macstadium_infrastructure/dhcp-server.tf
```diff
@@ -0,0 +1,50 @@
...
+    agent = true
+  }
+}
+
+resource "null_resource" "dhcp_server" {
+  triggers {
+    host_id = "${vsphere_virtual_machine.dhcp_server.uuid}"
+  }
+
+  connection {
```
meatballhat: Same here; is this `connection` used?

modules/macstadium_infrastructure/dhcp-server.tf
emdantrim: yep! this connection is used by the `macstadium_dhcp` module's `remote-exec`.

modules/macstadium_infrastructure/dhcp-server.tf
emdantrim: I... think? This is copied from the `macstadium_go_worker` module, and I assumed it was put in there for a good reason. :/

emdantrim:
Yep, it works!

modules/macstadium_infrastructure/dhcp-server.tf
meatballhat: How? :grin:  I thought `macstadium_dhcp` was having ssh user and host injected as variables.

modules/macstadium_infrastructure/dhcp-server.tf
meatballhat: I don't see anything using it, so ✂️ 🎉 !

meatballhat: Come to think of it, I don't see what this `resource "null_resource"` is doing at all 😁

PR #327, opened 2018-01-22
GCE worker updates

meatballhat:

The current (unused) systemd service definition for `travis-worker-{a,b,c}` on GCE does not work, as the `travis-worker-wrapper` script is not given a variable positional argument from which a container name is derived.

Use a systemd instance-style service definition so that stuff works right.  Also use the latest xenial tfw image.

## How can you test this?

- [x] delete a staging GCE worker host
- [x] `make plan`, `make apply`
- [x] cycle remaining hosts in gce-staging-1
- [x] cycle hosts in gce-production-1
- [x] cycle hosts in gce-production-2
- [x] cycle hosts in gce-production-3
- [x] cycle hosts in gce-production-4
- [x] cycle hosts in gce-production-5

### DIFF
--- a/assets/travis-worker/travis-worker@.service
+++ b/assets/travis-worker/travis-worker@.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Travis Worker
+After=docker.service
+Requires=docker.service
+
+[Service]
+ExecStart=/usr/local/bin/travis-worker-wrapper travis-worker-%I
+ExecStopPost=/bin/sleep 5
+Restart=always
+WorkingDirectory=/
+User=travis
+
+[Install]
+WantedBy=multi-user.target
--- a/gce-production-1/main.tf
+++ b/gce-production-1/main.tf
@@ -3,14 +3,14 @@ variable "env" {
 }
 
 variable "gce_bastion_image" {
-  default = "eco-emissary-99515/bastion-1478778272"
+  default = "eco-emissary-99515/bastion-1496867305"
 }
 
 variable "gce_gcloud_zone" {}
 variable "gce_heroku_org" {}
 
 variable "gce_worker_image" {
-  default = "eco-emissary-99515/tfw-1499625597"
+  default = "eco-emissary-99515/tfw-1516675156-0b5be43"
 }
 
 variable "github_users" {}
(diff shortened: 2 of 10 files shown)


gce-production-1/main.tf
```diff
@@ -63,8 +63,8 @@ module "gce_project_1" {
   worker_account_json_com       = "${file("${path.module}/config/gce-workers-production-1.json")}"
   worker_account_json_org       = "${file("${path.module}/config/gce-workers-production-1.json")}"
   worker_image                  = "${var.gce_worker_image}"
-  worker_instance_count_com     = 36
-  worker_instance_count_org     = 40
+  worker_instance_count_com     = 16
```
bogdanap: Is this change intentionally included here? (just asking because it wasn't mentioned in the description)

modules/gce_project/main.tf
```diff
@@ -236,7 +236,7 @@ resource "google_compute_firewall" "deny_target_ip" {
   project = "${var.project}"
 
   # highest priority
-  priority = "0"
+  priority = "1000"
```
bogdanap: The [docs](https://www.terraform.io/docs/providers/google/r/compute_firewall.html#priority) mention that 0 is the highest priority and 1000 is the default. What are we achieving with this change?

assets/travis-worker/travis-worker@.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
```
bogdanap: Do we still need the previous `worker.service` file then or does this replace it?

meatballhat: We don't need the previous worker service file on GCE, but it _is_ in use elsewhere 👍

gce-production-1/main.tf
meatballhat: Ah, thank you for bringing this up!  There's a bit in the commit message, but I forgot to update the PR description.  I made this change to reflect more closely the counts of hosts that were running in gce-production-1 at the time.  The majority of gce-production-1 hosts were in a `TERMINATED` state due to actions that I took in late 2017 prior to holiday time code/config freeze.

modules/gce_project/main.tf
meatballhat: This is another one that I addressed via commit comment but forgot to update the PR description 😅.  The documented highest priority of `0` appears to be either untrue or perhaps terraform is doing something special with the zero value such that repeat terraform plans wanted to change the value to `1000`.

PR #334, opened 2018-01-30
Move macstadium config out of keychain

emdantrim:
A lot of our macstadium application configuration is in keychain, which makes it really annoying to do minor config changes, which means we end up just not putting them in git at all.

Start moving application configs into terraform, so that we can interact with them in a more friendly way.

## How can you test this?
- [x] Compare new generated configs to old configs and make sure that they are similar enough
- [x] Make sure that configured applications are happy with their new workflow

Does all of this make sense? Is this actually less painful? Do you like this idea?


macstadium-pod-2/main.tf
```diff
@@ -453,8 +455,8 @@ module "collectd-vsphere-common" {
   librato_token                           = "${var.librato_token}"
   env                                     = "common"
   index                                   = "${var.index}"
-  collectd_vsphere_collectd_network_user  = "${var.collectd_vsphere_collectd_network_user}"
-  collectd_vsphere_collectd_network_token = "${var.collectd_vsphere_collectd_network_token}"
+  collectd_vsphere_collectd_network_user  = "collectd-vsphere-2"
```
Lyoness: var name needs to be declared since we're generating here, right?

macstadium-pod-1/Makefile
```diff
@@ -14,15 +14,10 @@ CONFIG_FILES := \
...
-		config/travis-worker-production-org-2 \
-		config/travis-worker-custom-1 \
-		config/travis-worker-custom-2 \
-		config/travis-worker-custom-4 \
-		config/travis-worker-custom-5 \
-		config/travis-worker-custom-6 \
+		config/travis-worker-production-com-common \
+		config/travis-worker-production-org-common \
+		config/travis-worker-staging-com-common \
+		config/travis-worker-staging-org-common \
```
Lyoness: dope!

macstadium-pod-1/workers.tf
```diff
@@ -0,0 +1,273 @@
...
+  ssh_user    = "${var.ssh_user}"
+  version     = "${var.travis_worker_version}"
+  env         = "production-com-2"
+  index       = "${var.index}"
+  worker_base_config = "${data.template_file.worker_config_common.rendered}"
+  worker_env_config  = "${file("${path.module}/config/travis-worker-production-com-common")}"
+
+  worker_local_config = <<EOF
+export TRAVIS_WORKER_TRAVIS_SITE="com"
+export TRAVIS_WORKER_POOL_SIZE="110"
```
Lyoness: 🙀 💯

modules/macstadium_go_worker/worker.tf
```diff
@@ -57,34 +45,26 @@ resource "null_resource" "worker" {
   }
 
   provisioner "file" {
-    content     = "${var.config_path}\n${data.template_file.worker_config.rendered}"
-    destination = "/tmp/etc-default-travis-worker-${var.env}"
+    content     = "${data.template_file.worker_upstart.rendered}"
+    destination = "/tmp/init-travis-worker-${var.env}.conf"
```
Lyoness: 👍

macstadium-pod-1/workers.tf
```diff
@@ -0,0 +1,272 @@
...
+  index              = "${var.index}"
+  worker_base_config = "${data.template_file.worker_config_common.rendered}"
+  worker_env_config  = "${file("${path.module}/config/travis-worker-production-org-common")}"
+
+  worker_local_config = <<EOF
+export TRAVIS_WORKER_TRAVIS_SITE="org"
+export TRAVIS_WORKER_POOL_SIZE="210"
+export TRAVIS_WORKER_PPROF_PORT="7070"
+export TRAVIS_WORKER_JUPITERBRAIN_ENDPOINT=http://${random_id.jupiter_brain_production_org_token.hex}@127.0.0.1:8081/
+export TRAVIS_WORKER_LIBRATO_SOURCE="travis-worker-production-org-macstadium-$${index}-1-dc18"
```
meatballhat: Based on the other code in this file, I'm guessing this should contain `${var.index}` and not `$${index}`.

macstadium-pod-2/main.tf
meatballhat: ⬆️ @Lyoness I don't follow 😁   Help!

modules/collectd_vsphere/collectd_vsphere.tf
```diff
@@ -62,16 +62,18 @@ data "template_file" "collectd_vsphere_upstart" {
...
-    collectd_network_token_file_signature = "${sha256(data.template_file.collectd_network_token_file.rendered)}"
-    collectd_snmp_config_signature        = "${sha256(data.template_file.snmp_conf.rendered)}"
-    name                                  = "${var.env}-${var.index}"
-    host_id                               = "${var.host_id}"
+    version                                              = "${var.version}"
+    config_signature                                     = "${sha256(file(var.config_path))}"
+    install_script_signature                             = "${sha256(data.template_file.collectd_vsphere_install.rendered)}"
+    librato_creds_signature                              = "${sha256(data.template_file.librato_conf.rendered)}"
+    collectd_config_signature                            = "${sha256(data.template_file.collectd_conf.rendered)}"
+    collectd_vsphere_collectd_network_username_signature = "${sha256(var.collectd_vsphere_collectd_network_user)}"
```
meatballhat: The collectd network username is no longer a secret, so does that mean that it should be used directly instead of its' `sha256` value?

emdantrim: good catch!

macstadium-pod-2/workers.tf
```diff
@@ -0,0 +1,260 @@
...
+  env                = "production-com-1"
+  index              = "${var.index}"
+  worker_base_config = "${data.template_file.worker_config_common.rendered}"
+  worker_env_config  = "${file("${path.module}/config/travis-worker-production-com-common")}"
+
+  worker_local_config = <<EOF
+export TRAVIS_WORKER_HARD_TIMEOUT=120m
+export TRAVIS_WORKER_TRAVIS_SITE="com"
+export TRAVIS_WORKER_POOL_SIZE="55"
+export TRAVIS_WORKER_JUPITERBRAIN_ENDPOINT=http://${random_id.jupiter_brain_production_com_token.hex}@127.0.0.1:8083/
```
soulshake: missing quotes here?

macstadium-pod-2/workers.tf
```diff
@@ -0,0 +1,260 @@
...
+  env                = "production-com-2"
+  index              = "${var.index}"
+  worker_base_config = "${data.template_file.worker_config_common.rendered}"
+  worker_env_config  = "${file("${path.module}/config/travis-worker-production-com-common")}"
+
+  worker_local_config = <<EOF
+export TRAVIS_WORKER_HARD_TIMEOUT=120m
+export TRAVIS_WORKER_TRAVIS_SITE="com"
+export TRAVIS_WORKER_POOL_SIZE="55"
+export TRAVIS_WORKER_JUPITERBRAIN_ENDPOINT=http://${random_id.jupiter_brain_production_com_token.hex}@127.0.0.1:8083/
```
soulshake: and here?

PR #339, opened 2018-02-12
Adjust values to newer terraform defaults

meatballhat:

Running `make plan` with Terraform v0.11.x results in unexpected changes, including re-creating all firewall rules that do not explicitly set `direction` and re-creating the bastion host due to mismatched image specification.

- Update the resource values to match existing state so that no changes are planned.
- Update the latest GCE image generation script to specify "self-link" values instead of "{project}/{name}"

## How can you test this?

Run a `make plan` in any of the `gce-production-.*` dirs.

### DIFF
--- a/bin/generate-latest-gce-images
+++ b/bin/generate-latest-gce-images
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
+require 'json'
 
 DEFAULT_GCE_PROJECT_ID = 'eco-emissary-99515'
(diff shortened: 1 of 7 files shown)


bogdanap:
Uh-oh, did I mess-up some of the projects with my applies earlier today? 🙀 (I don't remember seeing anything related to these changes in the plan, but I might have just missed it)

meatballhat:
@bogdanap I don't know! 😺  No worries at all 💖

PR #344, opened 2018-02-20
Add a connection-tracking service for use on NAT host

meatballhat:

The current GCE NAT implementation does not have a way to detect hosts that are creating "too many" connections.

Parse events from `conntrack -o xml -E conntrack` and log whenever a threshold is breached.  Future planned integration will combine this log output with a `fail2ban` action to temporarily jail the matching IP address.

## How can you test this?

Shame on me!  :fire: 

Should I move this into its own repository now, or now?

### DIFF
--- a/assets/nat/travis-nat-conntracker
+++ b/assets/nat/travis-nat-conntracker
@@ -1,0 +1,305 @@
+#!/usr/bin/env python
+import argparse
+import logging
+import sys
+import time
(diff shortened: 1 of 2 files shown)


meatballhat:
I'm closing this out in favor of continuing development + feedback over at https://github.com/travis-ci/nat-conntracker.

PR #346, opened 2018-02-21
Drop outbound internal traffic and to AWS metadata endpoint

soulshake:
- Drop requests to the AWS metadata endpoint
- Drop outbound container traffic to internal IPs

Addresses https://github.com/travis-pro/security/issues/65

Ideally this will be tested via https://github.com/travis-pro/nmap-check/pull/2
(which is passing right now, but I think it shouldn't be!)


modules/aws_asg/cloud-init.bash
```diff
@@ -44,8 +45,17 @@ main() {
...
   set +o errexit
+  # The loop of commands below drops any in-container traffic (which goes
+  # through the -I DOCKER chain) that attempts to talk to the docker registry host(s),
+  # and also drops anything destined for 128.0.0.0/16 (plus other reserved ranges)
+  reserved_ranges="
+    128.0.0.0/16
+  "
   dig +short "${registry_hostname}" | while read -r ipv4; do
     iptables -I DOCKER -s "${ipv4}" -j DROP || true
+    for r in $reserved_ranges; do
```
meatballhat: I prefer real arrays whenever possible :grin:

``` bash
reserved_ranges=(
  128.0.0.0/16
)
# ...
  for r in "${reserved_ranges[@]}"; do
# ...
```

modules/aws_asg/cloud-init.bash
```diff
@@ -44,8 +45,17 @@ main() {
...
+      iptables -I DOCKER -d "$r" -j DROP || true
```
meatballhat: `"${varname}"` style whenever possible :heart: &lt;/nit&gt;

soulshake:
This: 
```
  # The command below drops any requests to the AWS metadata API.
  iptables -t nat -I PREROUTING -p tcp -d '169.254.169.254' \
    --dport 80 -j DROP
```

results in this error:
```
iptables v1.6.0:
The "nat" table is not intended for filtering, the use of DROP is therefore inhibited.

Try `iptables -h' or 'iptables --help' for more information.
```

meatballhat:
Do we get the desired results when operating on the `filter` table?

soulshake:
I've replaced the aforementioned line with:
```
iptables -I FORWARD -d 169.254.169.254 -j REJECT
```

PR #361, opened 2018-03-03
Migrate gce-production-5 to separate graphs for network and workers

meatballhat:

The GCE production projects are not yet using the latest NAT goodies, as is currently the case for the GCE staging project.

Migrate `gce-production-5` to use separate graphs for network (`gce-production-net-5`) and workers (using different modules) as with the GCE staging project (`gce-staging-1` + `gce-staging-net-1`).

## How can you test this?

- [x] `make plan` && `make apply` in both `gce-production-net-5` and `gce-production-5`

These steps succeeded only after running the migration targets defined in this PR (`make export-net`, `make export-workers`, `make import-net`, `make import-workers`).


meatballhat:
**NOTE** this has already been applied in production

PR #362, opened 2018-03-05
Specify key pair 'aj' on ASG instances

soulshake:

Sometimes, instances are not able to complete cloud-init, which leaves them inaccessible.

Add my own key pair to the ASG launch configuration.

## How can you test this?

`make apply` in staging.

I don't know a way to add more than one key pair! :(

### DIFF
--- a/modules/aws_asg/main.tf
+++ b/modules/aws_asg/main.tf
@@ -217,6 +217,7 @@ resource "aws_launch_configuration" "workers" {
   name_prefix       = "${var.env}-${var.index}-workers-${var.site}-"
   image_id          = "${var.worker_ami}"
   instance_type     = "${var.worker_instance_type}"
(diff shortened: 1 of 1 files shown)


modules/aws_asg/main.tf
```diff
@@ -217,6 +217,7 @@ resource "aws_launch_configuration" "workers" {
   name_prefix       = "${var.env}-${var.index}-workers-${var.site}-"
   image_id          = "${var.worker_ami}"
   instance_type     = "${var.worker_instance_type}"
+  key_name          = "aj"
```
meatballhat: I think I would feel better about this being an injected variable :grin:, maybe with a default that's an empty string?

soulshake: I had a feeling you'd say that :D
Coming right up.

PR #364, opened 2018-03-05
Add cron job to monitor job CPU usage

soulshake:

We're not currently gathering any statistics on CPU usage by individual jobs. This is a start in the direction of doing so!

This PR adds a cron job to check container CPU usage via `docker stats` and report it to syslog in a format that's consumable by [honeytail](https://github.com/travis-ci/travis-honeytail).

## How can you test this?

Apply in staging, run some intensive jobs.

<strike>Is once per minute too frequent? I have a feeling it will be; once per hour might be better.</strike> The cron will run every 5 minutes.


modules/aws_asg/cloud-config.yml.tpl
```diff
@@ -86,6 +86,16 @@ write_files:
...
   permissions: '0644'
+- content: '${base64encode(file("${here}/high-cpu-check.bash"))}'
+  encoding: b64
+  owner: 'root:root'
+  path: /var/tmp/travis-run.d/high-cpu-check
+  permissions: '0750'
+- content: '${base64encode(file("${here}/high-cpu-check-crontab"))}'
+  encoding: b64
+  owner: 'root:root'
+  path: /etc/cron.d/kill-old-containers-crontab
```
meatballhat: Should this be `/etc/cron.d/high-cpu-check-crontab`?

modules/aws_asg/high-cpu-check.bash
```diff
@@ -0,0 +1,70 @@
...
+report_greedy_containers() {
+  max_cpu="$1"
+  : "${RUNDIR:=/var/tmp/travis-run.d}"
+  instance_id="$(cat "${RUNDIR}/instance-id")"
+  instance_ip="$(cat "${RUNDIR}/instance-ipv4")"
+
+  IFS=$'\n'
+  cpu_usage="$(docker stats --no-stream --format "{{.Container}} {{.CPUPerc}} {{.Name}}")"
+
+  for line in $cpu_usage; do
```
meatballhat: I would remove this intermediate step of pulling all of the `docker stats` output into a string, given that you already have a `| while read -r cid usage_as_float name` below.

modules/aws_asg/high-cpu-check.bash
```diff
@@ -0,0 +1,70 @@
...
+      if [ "$usage_as_int" -gt "${max_cpu}" ]; then
+        echo "$cid $usage_as_int $name"
+        __logger "warning" \
+          "high cpu usage detected" \
+          "status=noop" \
+          "instance_id=${instance_id}" \
+          "instance_ip=${instance_ip}" \
+          "cid=$cid" \
+          "cpu_usage=$usage_as_float" \
+          "name=$name"
```
meatballhat: Inconsistent variable dereferencing syntax `$foo` vs. `${foo}` (the latter is my preference!)  &lt;/nitpickery&gt;

modules/aws_asg/high-cpu-check.bash
```diff
@@ -0,0 +1,70 @@
...
+main() {
+  # shellcheck disable=SC2153
+  : "${MAX_CPU:=12000}"
+
+  local max_cpu
+  max_cpu="${MAX_CPU}"
+
+  {
+    IFS=$(echo -en "\n\b")
+    for greedy_cid in $(report_greedy_containers "${max_cpu}"); do
```
meatballhat: Likewise here, instead of declaring an `IFS` and slurping the subshell into a string (or, in this case, assuming it can be treated as an iterable), I def prefer using a `report_greedy_containers "${max_cpu}" | while read -r greedy_cid; do` :heart:

modules/aws_asg/high-cpu-check.bash
```diff
@@ -0,0 +1,76 @@
...
+
+  if [ -z "${cpu_usage}" ]; then
+    __logger "info" \
+      "no containers running" \
+      "status=noop" \
+      "instance_id=${instance_id}" \
+      "instance_ip=${instance_ip}"
+  fi
+
+  for line in ${cpu_usage}; do
```
meatballhat: Can this `for` loop and the zero-result check above be changed into a counter so that only the inner `while read -r ...` is needed and the zero result warning can happen afterward?

soulshake: Due to the fact that modifications to the variable don't persist outside the `while read` loop, I've removed the counter entirely, as I don't think it's really needed anyway.

modules/aws_asg/high-cpu-check.bash
```diff
@@ -0,0 +1,74 @@
...
+report_greedy_containers() {
+  max_cpu="${1}"
+  : "${RUNDIR:=/var/tmp/travis-run.d}"
+  instance_id="$(cat "${RUNDIR}/instance-id")"
+  instance_ip="$(cat "${RUNDIR}/instance-ipv4")"
+
+  IFS=$'\n'
+  stats="$(docker stats --no-stream --format "{{.Container}} {{.CPUPerc}} {{.Name}}" | tr -d '%')"
+
+  echo "${stats}" | while IFS=" " read -r cid usage_as_float name; do
```
meatballhat: :grin: I don't understand why the two-step of fetching stats and then `echo`'ing stats is needed :heart: 

Does this not work?

```
docker stats --no-stream --format "{{.Container}} {{.CPUPerc}} {{.Name}}" | while read -r cid usage_as_float name; do
  usage_as_float="${usage_as_float//%/}"
  # ... and everything else can stay as-is, AFAICT
```

modules/aws_asg/high-cpu-check.bash
```diff
@@ -0,0 +1,74 @@
...
+    "cron finished" \
+    "status=${status}"
+  exit "${code}"
+}
+
+report_greedy_containers() {
+  max_cpu="${1}"
+  : "${RUNDIR:=/var/tmp/travis-run.d}"
+  instance_id="$(cat "${RUNDIR}/instance-id")"
+  instance_ip="$(cat "${RUNDIR}/instance-ipv4")"
```
meatballhat: I'm sorry for not mentioning this earlier, but I think `max_cpu`, `instance_id`, and `instance_ip` should all be declared as `local`.

modules/aws_asg/high-cpu-check.bash
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+set -e
+set -o pipefail
+
+__logger() {
+  level="${1}" && shift
+  msg="${1}" && shift
+  date="$(date --iso-8601=seconds)"
+  log_msg="tag=cron time=${date} level=${level} msg=\"${msg}\""
```
meatballhat: Ack, sorry again!  Please declare `level`, `msg`, `date`, and `log_msg` as `local`, too :heart:

modules/aws_asg/high-cpu-check.bash
soulshake: Purely for readability!

meatballhat: :thinking: mmmm I'm sorry I'm stuck on this.  Defining a large-ish string variable and then `echo`'ing and, then checking if it's empty to indicate a counter ... It feels pretty brittle to me :confused: 

I understand if you want to get this merged as-is.  I'll stop commenting and get out of the way :heart:

PR #369, opened 2018-03-06
Migrate gce-production-4 to separate network and worker graphs

meatballhat:

The worker hosts in `gce-production-4` are not using a layer of NAT hosts as with `gce-production-5` and `gce-staging-1`.

Create a new `gce-production-net-4` terraform graph dir and migrate resources via helper scripts to/from `gce-production-4`.

## How can you test this?

`make plan` && `make apply` until it's green


meatballhat:
**NOTE** this is applied in production

PR #370, opened 2018-03-07
Migrate gce-production-3 to separate network and worker graphs

meatballhat:

The worker hosts in `gce-production-3` are not using a layer of NAT hosts as with `gce-production-5`, `gce-production-4`, and `gce-staging-1`.

Create a new `gce-production-net-3` terraform graph dir and migrate resources via helper scripts to/from `gce-production-3`.

## How can you test this?

`make plan` && `make apply` until it's green


meatballhat:
**NOTE** this has been applied in production

PR #373, opened 2018-03-07
Escape space instead of quotes

soulshake:

`cloud-init` is failing when trying to add instance-specific environment variables: 
```
travis-worker-wrapper[6572]: /var/tmp/travis-run.d/travis-worker.env: line 44: travis.ipv4:10.10.2.198: command not found
```

Escape the space instead of the surrounding quotes.

## How can you test this?

`make apply` in staging, then try for example:
```
docker ps --format '{{.Label "travis.instance_id"}}'
```

The instance ID label currently has a trailing `\`, e.g. `i-0419829205e303142\`. Any idea how to make that go away?


Lyoness:
>The instance ID label currently has a trailing \, e.g. i-0419829205e303142\. Any idea how to make that go away?

@soulshake 
#L24 has two backslashes `\\` wondering if both are needed or if only one is used for an escape and other prints the extra `\`

modules/aws_asg/cloud-init.bash
```diff
@@ -21,10 +21,10 @@ main() {
     if [[ -f "${RUNDIR}/instance-ipv4" ]]; then
       local ipv4_label
       ipv4_label="travis.ipv4:$(cat "${RUNDIR}/instance-ipv4")"
-      container_labels="${container_labels} ${ipv4_label}"
+      container_labels="${container_labels}\\ ${ipv4_label}"
```
Lyoness: I'm referring to this double backslash

soulshake: @Lyoness I tried with only one backslash, but that seems to result in the same behavior as pasted into the PR description unfortunately

igorwwwwwwwwwwwwwwwwwwww:
Hm, it looks like [this logic](https://github.com/travis-ci/terraform-config/blob/master/assets/bits/travis-combined-env#L24) from `travis-combined-env`, which is called from [travis-worker-wrapper](https://github.com/travis-ci/terraform-config/blob/master/assets/travis-worker/travis-worker-wrapper#L13) is stripping the quotes -- could that be the issue?

meatballhat:
@igorwwwwwwwwwwwwwwwwwwww FWIW the quote stripping is intentional, as the format required by Docker's `--env-file` option treats quotes as part of the value :heart:

modules/aws_asg/cloud-init.bash
meatballhat: Maybe this would be easier if we switched to the ","-delimited ":"-paired values like I'd thought it was supposed to be :sweat_smile:   I realllllly don't like multiple levels of escaping and I'd much rather avoid it if possible.

soulshake: @meatballhat I totally agree. Can we still merge this for now so that container labels are parsed correctly until worker is updated to look for comma-delimited labels?

soulshake: Update: I've opened [a PR](https://github.com/travis-ci/worker/pull/445) to address this in worker.

PR #374, opened 2018-03-08
Migrate gce-production-1 to separate network and worker graphs

meatballhat:

_TODO_

_TODO_

## How can you test this?

_TODO_


meatballhat:
**NOTE** this has been applied in production

PR #376, opened 2018-03-08
Create terraform graph dir for production DNS

meatballhat:

The documentation for which IP addresses we use in GCE is outdated, reporting a much larger range than we now use with NAT stuff in place.

Create a single DNS record that contains A records for all of the IP addresses assigned to production GCE NAT hosts, with the intent of listing this single record in the documentation.

## How can you test this?

- [x] `make plan` is happy

### DIFF
--- a/dns-production-0/Makefile
+++ b/dns-production-0/Makefile
@@ -1,4 +1,4 @@
-ENV := production-dns
+ENV := dns
 ENV_SHORT := production
 INDEX := 0
 
--- a/dns-production-0/Makefile
+++ b/dns-production-0/Makefile
@@ -8,11 +8,13 @@ PROD_TF_VERSION := v0.11.2
 TERRAFORM := $(HOME)/.cache/travis-terraform-config/terraform-$(PROD_TF_VERSION)
 
 .PHONY: .config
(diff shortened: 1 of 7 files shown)


dns-production-0/Makefile
```diff
@@ -0,0 +1,20 @@
+ENV := dns
+ENV_SHORT := production
+INDEX := 0
+
+include $(shell git rev-parse --show-toplevel)/terraform-common.mk
+
+PROD_TF_VERSION := v0.11.2
```
Lyoness: We've pinned terraform version to v0.9.11 because of the vpshere provider changes in the 0.10+ version change.

Lyoness: err. pinned for macstadium

Lyoness: Hardcoding the version might conflict with (bin/ensure-terraform)[https://github.com/travis-ci/terraform-config/blob/master/bin/ensure-terraform#L13] arch check ?

meatballhat: @Lyoness this version pin will only apply to this graph/dir :+1:

PR #377, opened 2018-03-09
Update bin/travis-worker-verify-config to check for presence of librato credentials

soulshake:

Librato credentials seem missing on worker hosts, and there doesn't seem to be a check for these.

I'm attempting to check for these via `bin/travis-worker-verify-config`.

## How can you test this?

Good question! A comment in `bin/travis-worker-verify-config` says it can be executed directly, but it seems to expect an argument:
```
    def initialize(argv: ARGV)
      @cloud_config_string_base64 = argv.fetch(0)
      @prog = File.basename($PROGRAM_NAME)
    end
```
And I can't figure out what it wants :/

Why aren't the Librato credentials being written to `/etc/default/travis-worker`?


meatballhat:
Won't be using this anymore! 😿 💔

PR #387, opened 2018-03-15
Add new enigma pool

soulshake:

We need a new pool

I have added a new pool

## How can you test this?

I don't think there's a way to test it per se.

What else needs to happen here before we apply?


aws-enigma-2/main.tf
```diff
@@ -0,0 +1,205 @@
...
+
+  config {
+    bucket         = "travis-terraform-state"
+    key            = "terraform-config/aws-shared-2.tfstate"
+    region         = "us-east-1"
+    dynamodb_table = "travis-terraform-state"
+  }
+}
+
+resource "random_id" "cyclist_token_org" {
```
meatballhat: Does this pool need cyclist, or can we run the ASG at a fixed scale?

aws-enigma-2/main.tf
```diff
@@ -0,0 +1,205 @@
+variable "aws_heroku_org" {}
+
+variable "env" {
+  default = "enigma"
```
meatballhat: I'm pretty sure this should be `"production"`.  Did you run into issues with it being that instead of `"enigma"`?

aws-enigma-2/main.tf
```diff
@@ -0,0 +1,205 @@
...
+module "aws_az_1e" {
+  source                    = "../modules/aws_workers_az"
+  az_group                  = "1e"
+  bastion_security_group_id = "${data.terraform_remote_state.vpc.bastion_security_group_1e_id}"
+  env                       = "${var.env}"
+  index                     = "${var.index}"
+  vpc_id                    = "${data.terraform_remote_state.vpc.vpc_id}"
+}
+
+module "aws_az_1e2" {
```
meatballhat: I think that the `aws_az_1*2` modules can be dropped.  Maybe even trim down to a single AZ?

aws-enigma-2/main.tf
```diff
@@ -0,0 +1,205 @@
...
+  syslog_address                         = "${var.syslog_address_org}"
+  worker_ami                             = "${data.aws_ami.tfw.id}"
+  worker_asg_max_size                    = 20
+  worker_asg_min_size                    = 1
+  worker_asg_namespace                   = "Travis/org"
+  worker_asg_scale_in_threshold          = 80
+  worker_asg_scale_in_evaluation_periods = 3
+  worker_asg_scale_in_period             = 300
+  worker_asg_scale_out_threshold         = 60
+  worker_asg_scale_out_qty               = 2
```
meatballhat: Yep, the more I look at this, I'm thinking this should be a fixed size ASG neutral or non-existent scaling policies.

aws-enigma-2/main.tf
```diff
@@ -0,0 +1,205 @@
...
+
+  provisioner "local-exec" {
+    command = <<EOF
+exec ${path.module}/../bin/format-images-json \
+  ${var.amethyst_image}=android,erlang,haskell,perl \
+  ${var.garnet_image}=default,go,jvm,node_js,php,python,ruby \
+  >${path.module}/generated-language-mapping.json
+EOF
+  }
+}
```
meatballhat: This `null_resource.language_mapping_json` can be removed, as we only use the one generated in `aws-production-2` :+1:

aws-enigma-2/worker.env
```diff
@@ -0,0 +1 @@
+../aws-production-2/worker.env
```
meatballhat: I prefer this to be a copy and not a symlink :heart:

aws-enigma-2/main.tf
```diff
@@ -0,0 +1,205 @@
...
+
+module "aws_asg_org" {
+  source                     = "../modules/aws_asg"
+  cyclist_auth_token         = "${random_id.cyclist_token_org.hex}"
+  cyclist_version            = "v0.5.0"
+  docker_storage_dm_basesize = "19G"
+  env                        = "${var.env}"
+  env_short                  = "${var.env}"
+  github_users               = "${var.github_users}"
+  heroku_org                 = "${var.aws_heroku_org}"
```
meatballhat: I'm assuming that if we omit a `cyclist` instance, that it will require some code changes in the `aws_asg` module, perhaps as something like `${var.cyclist_instance_count}` which defaults to `1` and may be set to `0`, which is then used as the `count` param value for any cyclist resources.

aws-enigma-2/main.tf
soulshake: Can we do both? I.e. have cyclist, but run at a fixed scale, like in staging?

meatballhat: If it's a fixed size, meaning there are no scaling policies, then there should be no scaling lifecycle events being published, which means that cyclist is sitting there doing nothing :grin:

soulshake: I was thinking it would be preferable to have a cyclist instance doing nothing (at least at first) than refactoring code so it works without cyclist, especially since we might want scaling at some point.
(This is partially driven by laziness, because I don't know how much work it would be to remove the dependency on cyclist, or to add it back later. And also partially a desire for parity among pools.)

meatballhat:
No longer needed 😿 💔

PR #416, opened 2018-04-11
Use container health scripts on Packet, too

meatballhat:

The container-based Linux environment on the Packet infrastructure is susceptible to the same or similar issues with container health that we see on EC2, but the tools are not in use.

Use the same tools over on Packet!

- [x] aws-staging-1 is still happy
- [x] aws-production-2 is still happy
- [x] packet-staging-1 is happy
- [x] packet-production-1 is happy


modules/aws_asg/cloud-config.yml.tpl
```diff
@@ -57,53 +57,45 @@ write_files:
...
   encoding: b64
   owner: 'root:root'
   path: /var/tmp/travis-run.d/check-docker-health
   permissions: '0750'
-- content: '${base64encode(file("${here}/check-docker-health-crontab"))}'
+- content: '${base64encode(file("${assets}/travis-worker/check-docker-health.crontab"))}'
   encoding: b64
   owner: 'root:root'
-  path: /etc/cron.d/check-docker-health-crontab
+  path: /etc/cron.d/check-docker-health.crontab
```
soulshake: I don't think crontab files can contain dots!

meatballhat: They sure can't!  I missed these :man_facepalming: Fixed now!

modules/aws_asg/cloud-config.yml.tpl
```diff
@@ -57,53 +57,45 @@ write_files:
...
+  path: /etc/cron.d/travis-worker-clean-up-containers
   permissions: '0644'
-- content: '${base64encode(file("${here}/high-cpu-check.bash"))}'
+- content: '${base64encode(file("${assets}/travis-worker/high-cpu-check.bash"))}'
   encoding: b64
   owner: 'root:root'
   path: /var/tmp/travis-run.d/high-cpu-check
   permissions: '0750'
-- content: '${base64encode(file("${here}/high-cpu-check-crontab"))}'
+- content: '${base64encode(file("${assets}/travis-worker/high-cpu-check.crontab"))}'
```
soulshake: Any reason not to just use the same filename in both places (i.e. in the repo and on the host)?

meatballhat: For me, at least, the `.crontab` extension enables syntax highlighting, which I was assuming at least one other person also appreciated :grin:

PR #418, opened 2018-04-18
Add rules to log and drop spoofed packets

emdantrim:
Allowing spoofed packets out into the internet is how we, collectively, as a species, get ants.

Drop packets that are coming into the local network's interface that aren't coming from the local subnet.

## How can you test this?
Deploy it and see if anything works, and then try spoofin' some packets! In fact, better yet, try spoofin' some packets now and see if they go through. (Though, that might be hard, because, well... spoofing.)

Is this necessary? Is this good? Do you like how I did it?


modules/packet_network/cloud-init.bash
```diff
@@ -101,6 +108,19 @@ __find_public_interface() {
...
 
+__find_local_interface() {
+  local iface=enp1s0f1
+  subnet="$(ip -o addr show | grep -v inet6 | grep -v " lo " | grep -v "bond" |
+    grep -v "docker" | awk '{ print $2}')"
+  echo "${iface:-enp1s0f1}"
+}
+
+__find_local_subnet() {
+  local subnet="192.168.0.1/24"
```
meatballhat: This default value for subnet should be `192.168.1.1/24`, IIUC :grimacing:

modules/packet_network/cloud-init.bash
```diff
@@ -92,6 +92,13 @@ __setup_networking() {
   if ! iptables -C FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT; then
     iptables -A FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
   fi
+
+  if ! iptables -C INPUT --in-interface "$(__find_local_interface)" --src ! "$(__find_local_subnet)" -j LOG --log-prefix "SPOOF"; then
+    iptables -A INPUT --in-interface "$(__find_local_interface)" --src ! "$(__find_local_subnet)" -j LOG --log-prefix "SPOOF"
+    if ! iptables -C INPUT --in-interface "$(__find_local_interface)" --src ! "$(__find_local_subnet)" -j DROP; then
+      iptables -A INPUT --in-interface "$(__find_local_interface)" --src ! "$(__find_local_subnet)" -j DROP
```
meatballhat: Given that the returns from `__find_local_interface` and `__find_local_subnet` are being used multiple times, what do you think about assigning them to local variables?

modules/packet_network/nat-dynamic-config.bash
```diff
@@ -103,6 +111,19 @@ __find_public_interface() {
...
+  local iface=enp1s0f1
+  subnet="$(ip -o addr show | grep -v inet6 | grep -v " lo " | grep -v "bond" |
+    grep -v "docker" | awk '{ print $2}')"
+  echo "${iface:-enp1s0f1}"
+}
+
+__find_local_subnet() {
+  local subnet="192.168.1.1/24"
+  subnet="$(ip -o addr show "$(__find_local_interface)" | awk '{ print $4}')"
+  echo "${subnet:-"192.168.0.1/24"}"
```
meatballhat: Should this line also have `192.168.1.1/24`?

PR #421, opened 2018-04-19
Bump Docker API version to match available

meatballhat:
given the currently-installed Docker version 18.04.0-ce


meatballhat:
This appears to be working well in production :+1:

PR #422, opened 2018-04-20
Increase inotify max instances for docker workers

joepvd:

Users occasionally run into not being able to create more file
listeners, as this is a shared resource on the docker host. This
fix sets fs.inotify.max_user_instances to WORKER_POOL_SIZE *
the default value.

Hopes are that these errors would plummet:
https://papertrailapp.com/groups/3719173/events?q=inotify_add_watch

Increase the max at image bootstrap time

## How can you test this?
Look at `/proc/sys/fs/inotify/max_user_instances`

Any.

### DIFF
--- a/modules/aws_asg/cloud-init.bash
+++ b/modules/aws_asg/cloud-init.bash
@@ -34,6 +34,8 @@ main() {
 
   __set_aio_max_nr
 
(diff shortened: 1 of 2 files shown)


meatballhat:
Won't be using this anymore 😿 💔

PR #423, opened 2018-04-23
Add planbeep target

joepvd:
The output of `make plan` gets often obfuscated by the large blurp of
`metadata.user-data`.

Pipe to filter.

## How can you test this?
`make planbeep`, confirm that Plan B keeps on beepin'

May I assume `GNU make`? :grin:


terraform-common.mk
```diff
@@ -67,6 +67,10 @@ console: announce
 plan: announce .config $(TRVS_TFVARS) $(TFSTATE)
 	$(TERRAFORM) plan -module-depth=-1 -out=$(TFPLAN)
 
+.PHONY: pplan
+pplan: announce .config $(TRVS_TFVARS) $(TFSTATE)
+	$(TERRAFORM) plan -module-depth=-1 -out=$(TFPLAN) | ruby -ne 'puts $$_.gsub(/(metadata.user-data: *")(.*)(".*)/, "\\1blurp\\3")'
```
meatballhat: FWIW, this is much the same reasoning behind the creation of the `plandiff` target, but I definitely understand how they differ.  I would prefer a slightly longer, more different than `plan` target name.  Maybe "planblurp" :sweat_smile: ?

soulshake:
That userdata blog is definitely super annoying when it appears! I wonder how hard it would be to decode it and show the actual diff someday?

I'm going to go ahead and counter @meatballhat's advice (sorry!) and suggest a more descriptive target name. `prettyplan`?

PR #431, opened 2018-04-25
Onboarding improvements

soulshake:

This PR is meant to highlight or address some of the [issues](https://github.com/travis-pro/pops/issues/747#issuecomment-384254405) that came up during @Halle's onboarding.

Only some of the problems are actually fixed by this PR, and the ones that are fixed might not be very polished.

I stopped when it occurred to me that this might be a good exercise for @Halle if she's interested :)


bin/pre-flight-checks
```diff
@@ -83,6 +118,7 @@ check_envvars() {
         GITHUB_USERNAME
         HEROKU_API_KEY
         TRAVIS_KEYCHAIN_DIR
+        SLACK_WEBHOOK
```
igorwwwwwwwwwwwwwwwwwwww: We should also get this one from the keychain automatically instead of requiring it to be copy pasted to `.env`.

terraform-common.mk
```diff
@@ -29,7 +29,7 @@ hello: announce
 
 .PHONY: .assert-ruby
 .assert-ruby:
-	@ruby -e "fail 'Ruby >= 2.3 required' unless RUBY_VERSION >= '2.3'"
+	@ruby -e "fail 'Ruby >= 2.3 required' unless RUBY_VERSION >= '2.4'"
```
igorwwwwwwwwwwwwwwwwwwww: The version in the error message should match the condition.

PR #432, opened 2018-04-25
Make high-cpu-check.bash report CPU usage of all jobs instead of just those using over 100%

soulshake:

The cronjob to check CPU usage currently only reports when a container-based job is using over 100% CPU.

The cronjob now reports any CPU usage over 0%.


assets/travis-worker/high-cpu-check.bash
```diff
@@ -70,14 +70,14 @@ report_greedy_containers() {
...
+  : "${MAX_CPU:=0}"
 
   local max_cpu
   max_cpu="${MAX_CPU}"
 
   {
-    report_greedy_containers "${max_cpu}" | while read -r greedy_cid; do
-      echo "greedy cid: ${greedy_cid}"
+    report_containers "${max_cpu}" | while read -r cid; do
+      echo "greedy cid: ${cid}"
```
meatballhat: Should this `echo`'d message also omit "greedy"?

igorwwwwwwwwwwwwwwwwwwww:
if this increases the log volume a lot, we may want to tweak the sample rate in travis-honeytail.

PR #442, opened 2018-05-09
Add pupcycler integration to Packet production

meatballhat:
As with #439 

- [x] production works well enough


Lyoness:
❇️ 
https://www.pinterest.com/pin/782219029000977357/

PR #444, opened 2018-05-09
Clean up job containers older than 60s without active exec

meatballhat:
- [x] works in production


assets/travis-worker/clean-up-containers.bash
```diff
@@ -37,13 +37,8 @@ main() {
...
-      not_killed_count="$((not_killed_count + 1))"
-      continue
-    fi
-
-    if ! __container_is_newer_than "${cid}" "${max_age}"; then
-      __logger info "${cid} is older than ${max_age}; removing! (name=${cn})"
+    if __container_must_be_cleaned "${cid}" "${max_age}"; then
+      __logger info "${cid} must be cleaned; removing! (name=${cn})"
 
       docker kill "${cid}" ||
```
soulshake: You're getting container IDs below via `docker ps --all` which will include any `Exited` containers that may be present. Running `docker kill` on such a container will result in an error like:
```
$ docker kill c0606705324d
Error response from daemon: Cannot kill container: c0606705324d: Container c0606705324de62755c03ae45e649cb745754607dea39f2e24239f8dbbe6309d is not running
```

Right?
I figure that might cause some confusion since it will result in the same error message (`failed to kill container name=${cn}`) as if it fails for some other reason. But maybe you're ok with that :)

assets/travis-worker/clean-up-containers.bash
```diff
@@ -135,6 +130,27 @@ EOF
...
+    __container_is_newer_than "${cid}" "${max_age}"; then
+    return 1
+  fi
+}
+
+__container_has_active_exec() {
+  local cid="${1}"
+  local exec_ids
+
+  exec_ids="$(docker inspect --format='{{ .ExecIDs | json }}' "${cid}")"
```
soulshake:
```
$ docker inspect --format='{{ .ExecIDs | json }}' $(docker ps -ql)
["351c339e17afa8afdebc493e33245923a98ce6c72b998ce427ffb632712b3583"]
```

Neat, I didn't know about this! I'm curious what (if anything) we can do with `ExecID`; I'm not finding much documentation on it at first glance. 

Is this a reliable way to distinguish containers that no longer have any business running??

assets/travis-worker/clean-up-containers.bash
meatballhat: Yep, it's definitely sloppy, but I mostly don't care (?) :grimacing:

assets/travis-worker/clean-up-containers.bash
meatballhat: My intent here was to use the presence of an `.ExecIDs` array to infer whether or not travis-worker is actively running the job, yep!

PR #456, opened 2018-05-28
Introduce canary asg

igorwwwwwwwwwwwwwwwwwwww:

Rolling out a new version of worker is quite risky, since deploys take a long time. One way to reduce that risk would be to allow performing a partial rollout on a small number of instances.

In the past we have performed such canary testing by manually changing the configuration on an existing instance. The goal of introducing a separate ASG is to make it easier to perform canary testing in a more repeatable way.

The canary ASG does not have any scaling rules, its size can be adjusted by setting the desired instances manually.

## How can you test this?

We can create the canary asg on staging and fiddle with the desired instances on the asg, as well as run a couple jobs on staging to make sure they are going through the canary workers.

Is this a viable and sufficiently maintainable approach?

refs https://github.com/travis-ci/reliability/issues/100

### DIFF
--- a/aws-production-2/main.tf
+++ b/aws-production-2/main.tf
@@ -414,3 +414,4 @@ module "aws_asg_cs50" {
 }
 
 # } cs50
+
--- a/aws-production-2/main.tf
+++ b/aws-production-2/main.tf
@@ -22,12 +22,21 @@ variable "worker_ami" {
   default = "ami-07dd1aada92124d1e"
 }
 
+variable "worker_ami_canary" {
+  # tfw 2018-03-08 00-09-18
+  default = "ami-07dd1aada92124d1e"
+}
+
 variable "amethyst_image" {
   default = "travisci/ci-amethyst:packer-1512508255-986baf0"
 }
 
 variable "garnet_image" {
   default = "travisci/ci-garnet:packer-1512502276-986baf0"
+}
+
+variable "worker_image_canary" {
+  default = "travisci/worker:v3.7.0-29-gf54cf99"
 }
 
 terraform {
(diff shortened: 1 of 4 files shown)


igorwwwwwwwwwwwwwwwwwwww:
Changes i've made:

* The current latest tfw ami is broken, so I've hardcoded a slightly older one for staging (instead of picking latest)
* The default staging asg now runs a stable worker version
* The canary asg runs the latest worker version
* The canary workers have a hostname like `i-0f5181c-staging-1-worker-org-ec2-canary`, which can be used for filtering in honeycomb and papertrail

igorwwwwwwwwwwwwwwwwwwww:
@soulshake just pointed out that this PR is almost a duplicate of #387 with some minor differences.

igorwwwwwwwwwwwwwwwwwwww:
> @igorwwwwwwwwwwwwwwwwwwww Can you expand on this, in a separate issue if you prefer?:

@meatballhat I've pinged you through a separate channel to discuss that part more. :)

aws-staging-1/main.tf
```diff
@@ -200,11 +206,51 @@ module "aws_asg_org" {
   worker_docker_image_php        = "${var.latest_docker_image_garnet}"
   worker_docker_image_python     = "${var.latest_docker_image_garnet}"
   worker_docker_image_ruby       = "${var.latest_docker_image_garnet}"
-  worker_docker_self_image       = "${var.latest_docker_image_worker}"
```
soulshake: I'm having a hard time seeing why this line gets taken out?

igorwwwwwwwwwwwwwwwwwwww: The main reason was to have the default staging ASG run the stable worker release, and run the latest one on the canary ASG only. The intention being to give us a more stable staging env, I'm happy to revert it though if folks think it's more valuable to always run latest, where latest means latest at time of last terraform apply.

aws-production-2/main.tf
```diff
@@ -261,6 +270,53 @@ module "aws_asg_org" {
   ]
 }
 
+module "aws_asg_org_canary" {
+  source                     = "../modules/aws_asg_canary"
+  cyclist_auth_token         = "${random_id.cyclist_token_org.hex}"
```
bogdanap: Do we still need to connect this to cyclist, since it won't be autoscaled?

bogdanap: Oh, never mind, I just noticed that it's not actually set further down. 😅

igorwwwwwwwwwwwwwwwwwwww: Cyclist is still needed to perform a graceful manual scale-in. We will manually bump the desired instances on the ASG. In the case of lowering desired instances (e.g. setting it to 0) we need cyclist to manage the lifecycle hooks, otherwise the instance will be terminated before the jobs are completed, which could impact production jobs running on the instance.

PR #472, opened 2018-06-14
Change hobby dynos to standard-1x

igorwwwwwwwwwwwwwwwwwwww:

Heroku API complains with error:

```
Heroku Enterprise accounts only support professional and private dynos.
```

Switch from hobby dynos to standard-1x.

## How can you test this?

terraform apply.


igorwwwwwwwwwwwwwwwwwwww:
Successfully applied to gce-staging-1, I'll go ahead and merge this.

PR #475, opened 2018-06-18
aws-staging-1: separate worker pool for free jobs on com (part 1)

igorwwwwwwwwwwwwwwwwwwww:
Part 1 of this change extracts cyclist out into its own module, but leaves the structure unchanged.

refs https://github.com/travis-ci/reliability/issues/116

PR #477, opened 2018-06-20
aws-staging-1: separate worker pool for free jobs on com (part 2)

igorwwwwwwwwwwwwwwwwwwww:
Follow-up to https://github.com/travis-ci/terraform-config/pull/475.

The current `aws_asg` module includes the `worker_queue` in some of the resource names, but not all of them. That makes it difficult to introduce new ASGs for other queues, as there are naming collisions. For cs50 we hack around this by setting `env` to `cs50-com` -- that's pretty icky though. So while that is one possible approach, I believe including the queue name explicitly to be a cleaner way forward.

I tried adding the `worker_queue` to disambiguate. This, however, requires re-creating the ASGs. In order to facilitate an incremental migration, I've copy-pasted the `aws_asg` module and applied the changes to the new copy. That way we can spin up the new ASGs and migrate traffic over slowly.

A more readable diff can be found [in this gist](https://gist.github.com/igorwwwwwwwwwwwwwwwwwwww/273cbaf5639b6e91255c4d2c878bdd6c).

The intention is to move everything to this new module and then replace `aws_asg`.

Does this seem like a good way forward? Or should we go with the env hack for now?

refs travis-ci/reliability#116

### DIFF
--- a/aws-staging-1/main.tf
+++ b/aws-staging-1/main.tf
@@ -191,7 +191,7 @@ module "aws_asg_com" {
 }
 
 module "aws_asg_com_free" {
-  source             = "../modules/aws_asg"
+  source             = "../modules/aws_asg_queue"
   cyclist_auth_token = "${random_id.cyclist_token_com.hex}"
   cyclist_url        = "${module.aws_cyclist_com.cyclist_url}"
   env                = "${var.env}"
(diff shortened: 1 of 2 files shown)


igorwwwwwwwwwwwwwwwwwwww:
This has been applied on staging.

igorwwwwwwwwwwwwwwwwwwww:
Booted a worker in staging, it created the `builds.ec2-free` queue, that queue has 3 consumers. 👍

PR #482, opened 2018-06-21
Pod-1 worker pool size tweak

Lyoness:
The change in https://github.com/travis-ci/packer-templates/pull/630
introduces a dedicated macpro to run packer builds for xcode build
images for the mac infra. We decided to provision this macpro from the
existing `Macpro_Pod_1` shared compute cluster.  As a result, worker
pool sizes need to change slightly so that the remaining macpros do not get
overencumbered during peak demand.

While this introduces other manual steps elsewhere in the packer build pipeline for xcode images
(see [this comment](https://github.com/travis-ci/packer-templates/pull/629#discussion_r197119741) ) that are the responsibility of the dev running the packer-build, this decision was made to alleviate the long feedback loops during the image making process due to CPU bottlenecks in smaller vms. 

## How can you test this?

Terraform plan output

### DIFF
--- a/macstadium-pod-1/workers.tf
+++ b/macstadium-pod-1/workers.tf
@@ -32,7 +32,7 @@ module "worker_production_org_1" {
 
   worker_local_config = <<EOF
 export TRAVIS_WORKER_TRAVIS_SITE="org"
-export TRAVIS_WORKER_POOL_SIZE="60"
+export TRAVIS_WORKER_POOL_SIZE="57"
 export TRAVIS_WORKER_PPROF_PORT="7070"
 export TRAVIS_WORKER_JUPITERBRAIN_ENDPOINT="http://${random_id.jupiter_brain_production_org_token.hex}@127.0.0.1:8081/"
 export TRAVIS_WORKER_LIBRATO_SOURCE="travis-worker-production-org-macstadium-${var.index}-1-dc18"
@@ -52,7 +52,7 @@ module "worker_production_org_2" {
 
   worker_local_config = <<EOF
 export TRAVIS_WORKER_TRAVIS_SITE="org"
-export TRAVIS_WORKER_POOL_SIZE="60"
+export TRAVIS_WORKER_POOL_SIZE="57"
 export TRAVIS_WORKER_PPROF_PORT="7071"
 export TRAVIS_WORKER_JUPITERBRAIN_ENDPOINT="http://${random_id.jupiter_brain_production_org_token.hex}@127.0.0.1:8081/"
 export TRAVIS_WORKER_LIBRATO_SOURCE="travis-worker-production-org-macstadium-${var.index}-2-dc18"
@@ -111,7 +111,7 @@ module "worker_production_com_1" {
   worker_local_config = <<EOF
 export TRAVIS_WORKER_HARD_TIMEOUT=120m
 export TRAVIS_WORKER_TRAVIS_SITE="com"
-export TRAVIS_WORKER_POOL_SIZE="40"
+export TRAVIS_WORKER_POOL_SIZE="37"
 export TRAVIS_WORKER_JUPITERBRAIN_ENDPOINT="http://${random_id.jupiter_brain_production_com_token.hex}@127.0.0.1:8083/"
 export TRAVIS_WORKER_LIBRATO_SOURCE="travis-worker-production-com-macstadium-${var.index}-1-dc18"
 EOF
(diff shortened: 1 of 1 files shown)


Halle:
Hi Carmen!

Our discussion including @mjm on Tuesday resulted in the decision that provisioning a Mac Pro for dedication to packer jobs had a prerequisite of having a painless method to switch the Mac Pro back into com at the end of the period of packer work, and vice versa for the rare times that new images need to be built, since otherwise that Mac Pro will mostly go to waste after this initial debug work on this script. From the other discussion (https://github.com/travis-ci/packer-templates/pull/629#discussion_r197111948) I understood that you didn't feel this was possible without a risk that was analogous to the full-infra risk discussed by Emma in that discussion, which is fine, but then that means that we can't fulfill this prerequisite and shouldn't re-provision a Mac Pro for this right now (the question of possible com overcapacity and what to do with it would be good to decouple from the question of how not to lock in that capacity's underuse via this dedicated application).

So, if we do still want to get a Mac Pro available for packer work, we would still first need a solution for the usage-switching question. I think this would be a great question for you and @mjm to put your heads together on and I believe that you will be able to solve it with an approach which does not represent new risk, and then this change in provisioning will be OK to proceed on.

IMO a solution could either look like something fully-automated that could be debugged as such, or it could be a manual process that is short and concisely-documented-enough that there is no conceptual barrier to its getting used at the beginning and end of every packer work period.

Thanks!

Lyoness:
Hi Halle, 

First, to answer the concerns on the move ahead with a temporary provision of packer macpro, I apologize, as these weren't addressed explicitly here. 

>the decision that provisioning a Mac Pro for dedication to packer jobs had a prerequisite of having a painless method to switch the Mac Pro back into com at the end of the period of packer work, and vice versa 

Dedicating a macpro temporarily to packer builds is easy to implement, as well as easy to revert, so the risk involved in making that infra change, and quickly, is minimal, so we don't run the risk of underuse.

I investigated risk in the _opposite_ direction (reduced capacity) and found potential near-term impact to be minimal/negligible: a temporary 3% reduction on total capacity for both org and com workers.  (1.8% capacity for org, 1.2% reduced capacity for com). 

Second, the realization that the addition of this PR seem to be an about-face of what was discussed/agreed on:

> We all need to have a talk about why agreements from the beginning of the week are consistently being dropped at the end of the week.

While this decision was made and implemented on the condition that the prerequisite had been determined to have been met, and also with [the understanding that we do so without making things permanent, and understanding the implications](https://github.com/travis-ci/packer-templates/pull/629#discussion_r197119741). This PR as well as [its packer-templates counterpart](https://github.com/travis-ci/packer-templates/pull/630) seem to negate that agreement, and present a "fully executed" feel to them.  My profuse apologies for the misunderstanding, that was not my intention, and the blunder was only seen in hindsight.  

Thank you bringing up the additional points about what needs to be considered for permanent implementation, they are spot-on and consideration up-front on them will save pain and suffering later. We'll catch up early next week to discuss further.  

Closing this PR and will revert [packer-templates/pull/630](https://github.com/travis-ci/packer-templates/pull/630)

Halle:
Hiya Carmen,

Thank you for this clear explanation – then it sounds like all we have to work on collectively is surfacing this info for each other so it's not hard to see that that preparatory steps we agree on in order to avoid creating tech debt are happening, and then we're good to go with making big positive changes like this one you've made possible here, without any downsides. I appreciate your taking the time and I'm excited to see what your change here will make possible for the packer work!

Halle:
OK, so the last step for this is just to document the method of switching this back and forth (even if it's very simple – in fact it's great if it's very simple and the doc is very short) and make sure that the team is agreed that this process has a single source of responsibility, i.e.:

1. When you want to work on Mac packer stuff, you can (and should!) "check out" the use of a Mac Pro if there is overcapacity which can be supported by data (supporting data has a healthy timeslice). Agreement on whether there is overcapacity should first be sought in combination with another teammate. It doesn't need universal consensus; any other teammate can help make this call as long as they are comfortable saying "no" if they have doubts that there is overcapacity. As well as being a gut check, a reason this shouldn't be a solo call is that it is possible that there are other simultaneous plans for capacity, so this is a chance to sync up. If there is no way to get agreement that there is overcapacity, we can discuss capacity! That's a good outcome.
2. With this agreement, the person who wants to work on packer is solely responsible for the full Mac Pro provisioning lifecycle: reprovisioning the machine so they can use it, and always returning it when done. This is to avoid long-term reprovisioning or tech-debt-y goings-on accidentally caused by diffusion of responsibility or high workload in another area. This will also demonstrate whether the process is bulletproof or if it needs automation or backstopping in some respect.

tl;dr: first get agreement that there is overcapacity, and then the person who works on the packer script is solely responsible for checking out the Mac Pro and checking it back in at the earliest reasonable opportunity. // @mjm @ryndaniels @emdantrim

Lyoness:
Done! PR here: 
https://github.com/travis-pro/manual/pull/842/files

Halle:
It's great! Thank you.

PR #485, opened 2018-06-22
add missing export to macstadium pod-2 TRAVIS_WORKER_SCRIPT_UPLOAD_TIMEOUT

igorwwwwwwwwwwwwwwwwwwww:
It is quite likely that that variable is not having any effect currently, and is falling back to worker's [default value](https://github.com/travis-ci/worker/blob/master/config/config.go#L30) of `3m30s`.


Lyoness:
Hi! I added the var in pod 1 as part of: https://github.com/travis-pro/team-blue/issues/835 investigation, but it didn't have any effect there, we may just want to omit that env var there if we want the values to be identical in both pods,  either way, thank you for catching the skew.

PR #489, opened 2018-06-27
Add resources for gcloud-cleanup serial output archiving

meatballhat:
as well as replacing the trvs-managed GCE credential JSON files with generated, terraform-managed credentials and policies.

### DIFF
--- a/modules/gce_worker_group/main.tf
+++ b/modules/gce_worker_group/main.tf
@@ -1,5 +1,8 @@
 variable "env" {}
-variable "gcloud_cleanup_account_json" {}
+
+variable "gcloud_cleanup_archive_retention_days" {
+  default = 8
+}
 
 variable "gcloud_cleanup_instance_filters" {
   default = "name eq ^(testing-gce|travis-job|packer-).*"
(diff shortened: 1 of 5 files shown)


modules/gce_worker_group/main.tf
```diff
@@ -82,6 +81,90 @@ module "gce_workers" {
...
+    }
+
+    condition {
+      age = "${var.gcloud_cleanup_archive_retention_days}"
+    }
+  }
+}
+
+resource "google_project_iam_custom_role" "gcloud_cleaner" {
+  role_id     = "gcloud_cleaner"
```
igorwwwwwwwwwwwwwwwwwwww: Was the `cleaner` vs `cleanup` here intentional?

modules/gce_worker_group/main.tf
```diff
@@ -82,6 +81,90 @@ module "gce_workers" {
...
+    "compute.regions.get",
+    "compute.regions.list",
+    "compute.zones.get",
+    "compute.zones.list",
+    "storage.objects.create",
+    "storage.objects.update",
+  ]
+}
+
+resource "google_service_account" "gcloud_cleanup" {
```
igorwwwwwwwwwwwwwwwwwwww: I'm a fan!

modules/gce_worker_group/main.tf
meatballhat: Yes, with my thinking being that this is a role as opposed to an identity, but I am happy to make it `cleanup` for consistency, too :+1:

PR #490, opened 2018-06-29
Add compute.globalOperations.list to permissions

joepvd:
Requeues at end of script due to:

```
msg="couldn't determine if instance was preempted" err="googleapi: Error 403: Required 'compute.globalOperations.list' permission for 'projects/travis-staging-1', forbidden" 
```

Add the permission that the API is complaining about.


meatballhat:
@joep I believe this will also need `compute.globalOperations.get`

PR #505, opened 2018-07-16
gce-production-1: create single com-free worker

igorwwwwwwwwwwwwwwwwwwww:
This is the minimal production rollout of the "com-free" worker pools. The staging version was https://github.com/travis-ci/reliability/issues/117.

This should be a non-invasive change and safe to apply.

We can test it by checking if the `builds.gce-free` queue gets created on com production rabbit.

refs https://github.com/travis-ci/reliability/issues/127


igorwwwwwwwwwwwwwwwwwwww:
Applied, and the `builds.gce-free` queue showed up in rabbit. Going to go ahead and merge this!

PR #506, opened 2018-07-16
aws-production-1: com-free worker pool

igorwwwwwwwwwwwwwwwwwwww:
This rolls out the `com-free` worker pool to production.

The instance count is set to 0 for the time being, since the large instances are a bit pricey. But that can be increased once we want to start taking traffic on that pool.

The patch should not change any production settings and should be safe to apply.

Once this has been applied we should be able to manually set the max and desired instance count to 1, see an instance boot, it should create the `builds.ec2-free` queue in rabbit. That indicates it's working.

refs https://github.com/travis-ci/reliability/issues/127


igorwwwwwwwwwwwwwwwwwwww:
Applying this, will then verify it in production.

igorwwwwwwwwwwwwwwwwwwww:
The instance came up successfully, and `builds.ec2-free` was created. Going to go ahead an merge! ✨

PR #511, opened 2018-07-20
Make tar file hash the same between runs

joepvd:

There have been unnecessary diffs caused by unimportant differences
between computers. This removes accidental differences from an archive.

> **Warning**: This introduces a dependency on GNU tar.

See https://reproducible-builds.org/docs/archives/ for some background.


igorwwwwwwwwwwwwwwwwwwww:
Hm, requiring GNU tar is not very osx friendly. It can be installed via homebrew via `brew install gnu-tar`, which aliases the binary to `gtar`.

It might be a good idea to detect OSX, try to locate `gtar` in that case, and if it's missing, provide instructions to install it.

joepvd:
Indirect evidence of its functioning: https://travis-ci.org/joepvd/travis-experiment/builds/4345396

igorwwwwwwwwwwwwwwwwwwww:
Tested locally on osx and it worked!

```make distclean plan
/Users/igor/code/terraform-config/terraform-common.mk:92: *** Please ensure GNU tar is installed and is available as gtar.  Stop.```

It would be nice to include something like `(e.g. brew install gnu-tar)` in that message.

PR #512, opened 2018-07-20
Create a dedicated host in pod-1 for building Packer images

mjm:

This is intended to fix a few issues that all stem from running macOS Packer builds on builders' dev machines:
* Mac builds take a long time (several hours), particularly from installing Xcode and simulators
* Mac builds have to maintain a connection to vCenter, which is unreliable when builds take so long
* Builders have to set up the right build environment (variables and tools) to be able to build correctly, which has gotten more complex recently

Create a dedicated `image-builder` host in pod-1 where we set up the environment needed to be able to run macOS packer builds:
* Go
* vsphere-images
* Packer
* The vSphere packer builder

We also set up tmux on the machine, which allows us to SSH in to the machine and start the build in a detachable session. There's a script installed on the machine to automate launching a properly configured tmux session. Another script will be included in packer-templates to rsync the contents of packer-templates to image-builder and then ssh in and run this script to start the build.

## How can you test this?

I've tested this by applying it to our current infrastructure, tainting the VM if needed to rebuild it from scratch. I then tried to start builds from it to verify that it had what it needed in the environment.

* Any thoughts on my Terraform work, especially as this is my first contribution to the Terraform configuration
* Any thoughts on this as an approach to the overall problem. I will say this isn't necessarily the end game here. It still involves us doing builds in a more manual way than I wish we were, just with a lot less pain. I want to keep thinking about how to make this a more resilient and automatic process.


Halle:
This is great, can we all four 'pair' and review it in action today as we did for the automation changes a couple weeks ago? I'd like to hear @emdantrim's thoughts on the Terraforming, and, with luck, also learn something, since this is still an area of our tooling I am getting to grips with.

PR #514, opened 2018-07-24
adding "openstack-production-1" plan

ghatwala:
Adding new plan for openstack infrastructure to create basic deployment of travis-worker with openstack backend for travis production. 

Tested over OSU setup

This just adds creation of workers which targets to listening to power queue's at travis production.


ghatwala:
@meatballhat - could u please let me know your review comments  ?

PR #518, opened 2018-08-07
Added "insecure" flag option to terraform.

ghatwala:
Added "inscure" flag option to terraform plan which was missed for openstack-docker-production

Used to fix scenarios wherein we use "--insecure" option to connect to openstack API's.

Able to validate successfully at PDC.

## How can you test this?
Able to validate successfully at PDC.


ghatwala:
@meatballhat  could u please review ?

PR #522, opened 2018-08-20
aws-staging-1: disable debug logging

igorwwwwwwwwwwwwwwwwwwww:

This is filling up the disk on staging. Here's where most of the disk space went:

```
i-0f0fd0e-staging-1-worker-com-ec2% ls -lah
total 5.2G
drwxrwxr-x  7 root   syslog 4.0K Aug 17 06:25 .
drwxr-xr-x 13 root   root   4.0K Apr 27 08:17 ..
...
-rw-r-----  1 syslog adm    2.6G Aug 20 13:43 daemon.log
...
-rw-r-----  1 syslog adm    744M Aug 20 13:43 syslog
-rw-r-----  1 syslog adm    1.9G Aug 17 06:25 syslog.1
...
```

Both of those files contain worker logs. The debug logging is producing excessive log volumes.

Yes we should probably also fix logrotate, but this is a good first step.


igorwwwwwwwwwwwwwwwwwwww:
Applied!

Issue #347, opened 2018-02-23
Use instance groups or k8s for GCE workers


meatballhat:
Instead of provisioning individual instances, create instance templates and instance groups (1 each for org and com?) and perhaps take advantage of rolling updates and health checks while we're at it.  OR get wacky with some k8s fun maybe?

meatballhat:
Done via #529!

PR #529, opened 2018-08-31
manage gce workers via instance groups

igorwwwwwwwwwwwwwwwwwwww:

Currently there is no managed way to cycle GCE worker instances. While we still need to figure out how to do graceful shutdowns, instance groups might make it more straight forward to roll out new worker versions and configs on GCE.

We'll need to migrate the traffic over to these worker instances before we shut down the old ones.

Instance groups, managed by google cloud.

## How can you test this?

We can apply it on staging and make sure the workers come up successfully.

What do y'all think about this approach?

refs https://github.com/travis-ci/reliability/issues/112
refs https://github.com/travis-ci/terraform-config/issues/347

### DIFF
--- a/gce-staging-1/main.tf
+++ b/gce-staging-1/main.tf
@@ -98,9 +98,13 @@ module "gce_worker_group" {
 
   worker_zones = "${var.worker_zones}"
 
-  worker_instance_count_com      = "${length(var.worker_zones)}"
-  worker_instance_count_com_free = "1"
-  worker_instance_count_org      = "${length(var.worker_zones)}"
+  worker_instance_count_com      = "0"
+  worker_instance_count_com_free = "0"
+  worker_instance_count_org      = "0"
+
+  worker_managed_instance_count_com      = "${length(var.worker_zones)}"
+  worker_managed_instance_count_com_free = "1"
+  worker_managed_instance_count_org      = "${length(var.worker_zones)}"
 
   worker_config_com = <<EOF
 ### worker.env
--- a/modules/gce_worker/main.tf
+++ b/modules/gce_worker/main.tf
@@ -300,7 +300,7 @@ resource "google_compute_instance_template" "worker_com_free" {
   name_prefix = "${var.env}-${var.index}-worker-com-free-template-"
 
   machine_type = "${var.machine_type}"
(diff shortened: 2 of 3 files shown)


igorwwwwwwwwwwwwwwwwwwww:
Not sure yet how to deal with this:

```
* module.gce_worker_group.module.gce_workers.google_compute_instance_template.worker_org[2] (destroy): 1 error(s) occurred:

* google_compute_instance_template.worker_org.2: Error deleting instance template: googleapi: Error 400: The instance_template resource 'projects/travis-staging-1/global/instanceTemplates/staging-1-worker-org-c-template-f48fca9' is already being used by 'projects/travis-staging-1/zones/us-central1-c/instanceGroupManagers/worker-org-c', resourceInUseByAnotherResource
```

igorwwwwwwwwwwwwwwwwwwww:
I used this to drain the project:

```
for g in worker-{org,com,com-free}; do
  for z in {a,b,c,f}; do
    gcloud compute instance-groups managed resize --size=0 $g-$z --zone=us-central1-$z --project travis-staging-1
  done
done
```

And then applied with `create_before_destroy = true`. It requires draining the whole project gracefully though.

igorwwwwwwwwwwwwwwwwwwww:
The instance template is missing the amqp config from `config/worker-com.env`, not sure yet what is up with that.

igorwwwwwwwwwwwwwwwwwwww:
One possible approach for incremental rollout would be to maintain multiple instance templates in terraform, and make them immutable. Might be a huge pain to manage though.

Considering the issues we've been having with delete instance templates with the NATs this it might actually be worthwhile to do this though.

igorwwwwwwwwwwwwwwwwwwww:
To cycle an instance:

```
gcloud beta compute instance-groups list-instances worker-org --project=travis-staging-1 --region=us-central1

gcloud beta compute instance-groups managed recreate-instances worker-org --instances=staging-1-worker-org-0j5x --project=travis-staging-1 --region=us-central1
```

igorwwwwwwwwwwwwwwwwwwww:
Tested this end-to-end on staging and it's working! 👍

I've introduced a separate variable for the managed instance count, so we can roll this out incrementally.

PR #537, opened 2018-09-05
First draft of separate staging environment for Mac infra

Halle:

We would like a separate staging environment for testing on the Mac infrastructure. This is the first draft of the first iteration towards that goal.

I've created a new graph for macstadium-staging and removed its content from macstadium-pod-1. and macstadium-pod-2. I have renamed macstadium-pod-1 and macstadium-pod-2 to macstadium-prod-1 and macstadium-prod-2 to indicate that they represent the production environments versus the staging environment. I've renamed version variable names like jupiter_brain_staging_version which reference staging since that will no longer be necessary, and I've set wjb to install the newest Ubuntu for staging.

## How can you test this?

I'm interested in starting with a visual review of the changes. It's my expectation that we'll discuss the changes and then revise rather than approve and merge.

Anything – I figure there is some content here to discuss. Thanks Matt!


macstadium-staging/main.tf
```diff
@@ -0,0 +1,149 @@
...
+variable "fw_snmp_community" {}
+variable "vsphere_user" {}
+variable "vsphere_password" {}
+variable "vsphere_server" {}
+variable "vsphere_ip" {}
+
+terraform {
+  backend "s3" {
+    bucket         = "travis-terraform-state"
+    key            = "terraform-config/macstadium-staging-hopethisworks-terraform.tfstate"
```
mjm: We can probably take out the "hopethisworks" from here. I think it's a weird relic from the struggles of bringing up the MacStadium terraform the first time.

macstadium-staging/main.tf
```diff
@@ -0,0 +1,149 @@
...
+  ssh_user = "${var.ssh_user}"
+
+  config = [
+    {
+      name               = "jupiter-brain-staging-org"
+      frontend_port      = "8082"
+      backend_port_blue  = "9082"
+      backend_port_green = "10082"
+    },
+  ]
```
mjm: I think we lost the "jupiter-brain-staging-com" entry here.

macstadium-staging/workers.tf
```diff
@@ -0,0 +1,142 @@
+variable "latest_travis_worker_version" {}
```
mjm: I think we should decide how we want the worker version to behave in staging.

In the existing configuration, the `travis_worker_version` variable was used for non-staging workers, and `latest_travis_worker_version` was used for staging workers. The latter variable is populated automatically by a script using the most-recent GitHub release of worker.

In my experience, while that may be nice for making sure staging is always running a recent worker, it's not really what I wanted and it made things clumsy when I wanted to test a new worker version on staging. I think I would prefer that we remove `latest_travis_worker_version`, and use the explicit `travis_worker_version` here for the staging workers, similar to how production works. This is even more feasible now, since the variable will be distinct for the two environments, so we can easily test a new version in staging.

Halle:
Let me know if this looks ready to go, thanks!

mjm:
This looks good to me. I'd like to review the Terraform plan, since this will mean taking things out of production.

mjm:
Hey friends! I'm taking over this change, and I've made some commits that I think get this ready to go. When deployed, this will bring remove workers and more that are for staging from the prod machines, and instead bring them up on the staging WJB machine. Would love another set of eyes on it.

The staging environment will use the same version of Terraform as other infras now. For the time being, macstadium production still uses v0.9.

Lyoness:
you can take out the threatstack keys for the staging module
https://github.com/travis-pro/team-blue/issues/836

PR #554, opened 2018-09-20
Use recreate-instances instead of rolling-update for GCE NATs

bogdanap:
https://github.com/travis-ci/reliability/issues/149

As stated in the title, use the [recreate-instances](https://cloud.google.com/sdk/gcloud/reference/compute/instance-groups/managed/recreate-instances) command to replace the NATs.

## How can you test this?
Tested on staging

All of the feedback!


bin/gcloud-recreate-nat-instances
```diff
@@ -16,38 +17,23 @@ def main
...
 
-  groups_templates = {}
-  templates.each do |t|
-    groups.each do |g|
-      groups_templates[g] = t if t =~ /.+#{g}-template-.+/
-    end
-  end
-
   groups_zones.each do |instance_group, zone|
+    instance_list = JSON.parse(`gcloud compute instance-groups list-instances #{instance_group} --zone=#{zone} --project=travis-staging-1 --format=json`)
```
igorwwwwwwwwwwwwwwwwwwww: project is hard coded here

bogdanap: oh, copy-pasta fail! 😅 thank you!

PR #557, opened 2018-09-20
Use correct source and tag arguments for google_compute_firewall resources

meatballhat:


meatballhat:
@igorwwwwwwwwwwwwwwwwwwww it hadn't yet, but definitely would if left unchanged :sweat_smile:

PR #565, opened 2018-10-02
set TRAVIS_WORKER_INFRA on ec2, gce, macstadium

igorwwwwwwwwwwwwwwwwwwww:
This change introduces an "infra" tag in worker which should be useful for propagating source information to downstream systems (e.g. travis-logs), so that they can compute metrics grouped by infra.

See https://github.com/travis-ci/worker/pull/508 for the worker change that actually makes use of this setting.

refs travis-ci/reliability#190


bogdanap:
While we're still using Packet, maybe it's worth adding it there too.

PR #568, opened 2018-10-04
gce-production: decrease pool size from 90 to 50

igorwwwwwwwwwwwwwwwwwwww:
To decrease risk of increased pool sizes, I want to decrease them while increasing the number of single-process worker nodes.

The Math(tm) for moving from pool size of 90 to 50 while preserving capacity:

```
90*15 = 1350 => 1350/50 = 27
90*18 = 1620 => 1620/50 = 32,4
```

refs travis-ci/reliability#193


meatballhat:
:tada:

PR #571, opened 2018-10-10
remove public network interface on warmer pool instances

igorwwwwwwwwwwwwwwwwwwww:
Worker creates job VMs without a public IP, so that all traffic goes through the NAT. This change ensures pre-warmed instances have the same behaviour.

refs travis-ci/reliability#195


igorwwwwwwwwwwwwwwwwwwww:
Confirmed working on staging!

PR #582, opened 2018-10-17
gce: bump worker to v4.5.1

igorwwwwwwwwwwwwwwwwwwww:


igorwwwwwwwwwwwwwwwwwwww:
I'm inclined  to apply & merge this, since it is working fine locally. We should figure out what is up with the GNU tar error though.

PR #583, opened 2018-10-17
gce: raise rate limit from 20 req/s to 30 req/s

igorwwwwwwwwwwwwwwwwwwww:

We are spending a lot of time in rate limiting.

Raise the rate limit to 30 req/s. We have quota for 3000 requests per 100 seconds. So this is cutting it pretty close. We may need to apply for an increase from GCP.

## How can you test this?

We can look at `HEATMAP(gce_api_rate_limit_ms)` in Honeycomb.


igorwwwwwwwwwwwwwwwwwwww:
We're getting the same GNU tar issue as in https://github.com/travis-ci/terraform-config/pull/582.

igorwwwwwwwwwwwwwwwwwwww:
This has been applied on production-1 and production-2.

PR #585, opened 2018-10-25
Move to Xenial image

joepvd:
Upgrade tar from 1.27.1 to 1.28, with which it gains the option `--sort`.


igorwwwwwwwwwwwwwwwwwwww:
OMG thank you! <3

PR #587, opened 2018-10-27
Use latest tfw bits + vanilla bionic on GCE

meatballhat:

The GCE way of running travis-worker is not taking advantage of the latest `tfw` goodness as was developed for use on Packet.

Use latest `tfw` and switch to a vanilla Ubuntu Bionic image.

## How can you test this?

- [x] it works in staging :tada:

WDYT? This would also mean that several things could be removed in the packer-templates, and maybe also travis-cookbooks repo(s).

Closes #516


igorwwwwwwwwwwwwwwwwwwww:
Generally huge 👍!

I I'll link [`travis-ci/worker-gcp-standalone`](https://github.com/travis-ci/worker-gcp-standalone) since it is sort of the barebones version (though it's missing stuff like ssh keys from github, papertrail logging).

PR #592, opened 2018-10-31
GCE service account sharding

meatballhat:
Further details available in https://github.com/travis-ci/team-build-env-linux/issues/77.

This has already been applied and tested in staging.

### DIFF
--- a/modules/gce_worker/main.tf
+++ b/modules/gce_worker/main.tf
@@ -55,21 +55,21 @@ resource "google_service_account" "workers" {
 
 resource "google_service_account" "workers_com" {
   count        = "${var.worker_service_accounts_count_com}"
-  account_id   = "workers-com-${lookup(var.regions_abbrev, var.region, "unk")}${count.index+1}"
+  account_id   = "workers-com-${lookup(var.regions_abbrev, var.region, "unk")}-${count.index+1}"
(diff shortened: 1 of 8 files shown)


modules/gce_worker/main.tf
```diff
@@ -23,6 +31,18 @@ variable "worker_image" {
   default = "ubuntu-os-cloud/ubuntu-1804-lts"
 }
 
+variable "worker_service_accounts_count_com" {
+  default = 4
```
Lyoness: any reason for 4 as default?  Just a number to begin with?

igorwwwwwwwwwwwwwwwwwwww:
This looks good to me!

We could also consider creating a first class concept of a shard and then have one instance group per shard, with a dedicated service account per group. That would make the worker process start more predictable, and allow us to use instance service accounts again. :)

It would also pave the way for moving more stuff to a single project, and allow us to run `worker-deploy` in smaller batches.

To be clear, all of the above is by no means a blocker, happy to move forward as-is and think about those things Later(tm). 👍

meatballhat:
I am extremely on board with a first class shard concept :saxophone: :bear:

modules/gce_worker/main.tf
meatballhat: My thinking was that it's the default `length(var.zones)` :man_shrugging:

PR #596, opened 2018-11-01
Replace macstadium-staging with a Kubernetes cluster

mjm:
I don't know how much of this will end up getting merged in, but I've been experimenting with provisioning a Kubernetes cluster in the MacStadium staging environment. We can chat about the approach here.


macstadium-staging/main.tf
```diff
@@ -72,99 +31,43 @@ provider "vsphere" {
...
-  ssh_host                = "${module.macstadium_infrastructure_staging.util_ip}"
-  ssh_user                = "${var.ssh_user}"
-  vsphere_monitor_version = "${var.vsphere_monitor_version}"
-  config_path             = "${path.module}/config/vsphere-monitor"
-  index                   = "${var.index}"
-}
+module "kubernetes_cluster" {
+  source                 = "../modules/macstadium_k8s_cluster"
+  name_prefix            = "cluster-staging"
+  ip_base                = 100
```
Lyoness: hey what is this ip_base value?  I looked in K8s terraform provider but no luck.

mjm: It's the starting point for the IPs we assign to the cluster VMs. So the master will be .100 and the nodes will be .101 and .102 in this case.

modules/macstadium_k8s_cluster/master.tf
```diff
@@ -0,0 +1,119 @@
...
+      ipv4_gateway    = "10.182.64.1"
+      dns_server_list = ["1.1.1.1", "1.0.0.1"]
+      dns_suffix_list = ["vsphere.local"]
+    }
+  }
+
+  wait_for_guest_net_routable = false
+
+  connection {
+    host  = "${vsphere_virtual_machine.master.clone.0.customize.0.network_interface.0.ipv4_address}"
```
Lyoness: what are all these `0`s  in the interpolation syntax? Attributes as `nil`?

mjm: The clone and customize sections are represented as arrays, so you have to index into the first item.

Lyoness:
Looks awesome @mjm 👍

PR #597, opened 2018-11-02
gce-production-3: resurrect, starting with NATs

igorwwwwwwwwwwwwwwwwwwww:
This is the beginning of resurrecting gce-production-3. This is partially to address the GCE API rate limiting issues we're having with our current quota, but also to allow rolling out worker in smaller batches.

* We need to provision the NATs first, so that we can announce the IPs. AFAIK we have a policy of giving a couple days notice for this kind of change.
* Next, we should make sure our quotas match what we have in the other prod projects.
* Then, we can start provisioning workers to that project, splitting the existing number of workers evenly between them.

### DIFF
--- a/gce-production-net-3/main.tf
+++ b/gce-production-net-3/main.tf
@@ -61,3 +61,32 @@ provider "google" {
 
 provider "aws" {}
 provider "heroku" {}
+
+module "gce_net" {
+  source = "../modules/gce_net"
+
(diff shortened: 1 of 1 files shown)


igorwwwwwwwwwwwwwwwwwwww:
This has been applied in `gce-production-net-3`. New NAT IPs are:

```
104.154.255.220
35.232.158.140
35.239.89.243
23.251.144.230
```

PR #599, opened 2018-11-06
HTTP proxy based build caching

meatballhat:

Specifically, the reliability of fetching resources from ppa.launchpad.net, but more generally adding a layer of caching for anything that's HTTP `GET` or `HEAD` request-able.

Run [squignix](https://github.com/tianon/squignix) because (see reasons listed in squignix README). ~Load balancing isn't included here because it's a [bit of a pain to set up](https://cloud.google.com/load-balancing/docs/internal/setting-up-internal), but we can add it if/when deemed necessary.~

## How can you test this?

It works via `http_proxy` env var:
- example job: https://staging.travis-ci.org/meatballhat/yolo-octo-adventure/jobs/731096
- squignix-side logs: https://gist.github.com/meatballhat/74f2dd22fb96ccffec2ec31aa107dc7b

It works via APT's `Acquire` config bits:
- example job: https://staging.travis-ci.org/meatballhat/yolo-octo-adventure/jobs/731328
- squignix-side logs: https://gist.github.com/meatballhat/8124b84903a50a6a2dc11d211849854c


build-caching-staging-1/cloud-init.bash
```diff
@@ -0,0 +1,64 @@
...
+    logger running setup substep="${substep}"
+    "__setup_${substep}"
+  done
+}
+
+__wait_for_docker() {
+  local i=0
+
+  while ! docker version; do
+    if [[ $i -gt 600 ]]; then
```
soulshake: Just curious--what's this about? (In what cases would `docker version` exit higher than `600`?)

build-caching-staging-1/main.tf
```diff
@@ -0,0 +1,204 @@
...
+  }
+}
+
+resource "google_compute_backend_service" "build_cache" {
+  name        = "build-cache-backend"
+  description = "backend servers for build cache"
+  port_name   = "http"
+  protocol    = "HTTP"
+  timeout_sec = 10
+  enable_cdn  = false
```
soulshake: Maybe we could try the CDN magic in the future! (unless there's a reason not to)

build-caching-staging-1/cloud-init.bash
meatballhat: @soulshake Ah, that's the 10m timeout counter comparison, not an exit code check 😁

build-caching-staging-1/main.tf
meatballhat: I'm happy to keep looking into this in the future, yep! I hesitated to glom it onto this work because of the problems associated with cache invalidation. The solution here is extremely simplified and doesn't deal with cache invalidation outside of capping the `tmpfs` size.

build-caching-staging-1/cloud-init.bash
soulshake: LOL sorry, reading FTW

PR #603, opened 2018-11-08
Decommission a single macstadium prod worker

mjm:

We want to slowly but surely migrate the MacStadium-based infra to run on Kubernetes. We have a cluster running, so now we can start moving things over, one worker at a time.

Pool size of `travis-worker-production-org-4` is reduced to zero so that we can gracefully bring down the worker before removing it entirely. If we didn't do this, we'd have to either stop it forcefully or it would just restart again after a graceful shutdown. So I've opted to remove its pool so at least it won't scoop up jobs when we gracefully shut it down.

I also added S3 users for build tracing for the Kubernetes-based workers.

## How can you test this?

Arguably, this is the test! This will be the process we use for other workers, and this is the first test of doing that on a small section of our infrastructure.

Any.

### DIFF
--- a/macstadium-prod-1-cluster/main.tf
+++ b/macstadium-prod-1-cluster/main.tf
@@ -97,3 +97,44 @@ output "imaged_secret_key" {
   value     = "${module.aws_iam_user_s3_imaged.secret}"
   sensitive = true
 }
+
+// These users are for the worker instances that will run on the cluster.
+// The credentials are outputs so they can be copied into the keychain.
+//
+// If the users ever get recreated, those credentials need to get copied
+// again so the Kubernetes secrets can be updated.
+//
+// This is not ideal, so I'd like to find a better way to manage this at
+// point.
+
+module "worker_com_s3_user" {
+  source         = "../modules/aws_iam_user_s3"
+  iam_user_name  = "worker-macstadium-prod-1-com"
+  s3_bucket_name = "build-trace.travis-ci.com"
+}
+
+output "worker_com_access_key" {
+  value     = "${module.worker_com_s3_user.id}"
+  sensitive = true
+}
+
+output "worker_com_secret_key" {
+  value     = "${module.worker_com_s3_user.secret}"
+  sensitive = true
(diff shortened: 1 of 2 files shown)


Lyoness:
Awesome!  
As you decommission one worker via pool size, will you have a companion PR that will follow the same structure as the current prod-1 graph directory?  Something like `macstadium-prod-1-cluster/workers.tf` where you'll add one at a time with the matching `${var.worker_org_pool_size}` that was decomissioned?  

We have good headroom to muck with capacity this way, a nice approach ✅

mjm:
Actually, another approach would be to just reduce this worker to pool size 20, and bring up a single worker in Kubernetes. We could do that across the board as we do the first half of these worker migrations, which would leave the process running on the existing infra, so if we need to go back to it for any reason, it's less disruptive to do so.

Lyoness:
>It's actually a little more interesting than that. New Kubernetes services won't be set up in Terraform: instead, they live here: https://github.com/travis-ci/kubernetes-config. 

Ah ok, glad I asked! So making changes to in mac-land will have a different workflow now, split by infra/ services? 

Infra config and setup (to the k8s cluster)== this repo
Services config (k8s) == kubernetes-config repo

>Actually, another approach would be to just reduce this worker to pool size 20, and bring up a single worker in Kubernetes. We could do that across the board as we do the first half of these worker migrations, which would leave the process running on the existing infra, so if we need to go back to it for any reason, it's less disruptive to do so.

This also is reasonable too; I also think this 2nd approach would reduce the blast radius of a downed worker should something go wrong.

mjm:
Yes, that's exactly right. Terraform just isn't all that good at managing Kubernetes services, from what I can tell (I tried at first). My research into how people are using Kubernetes suggests that this sort of split is common: there is a separation between management of the cluster and management of what runs on the cluster.

mjm:
And for now this split will only be true for Mac-land, but if Linux infra moves to Kubernetes on GKE later, that can use the same process and repos.

PR #604, opened 2018-11-09
gce-production-3: resurrect, capacity

igorwwwwwwwwwwwwwwwwwwww:
This patch restores capacity to the production-3 project. It does so by re-distributing the number of workers evenly between the three projects.

* Before we do this, we are having another chat about getting our API quotas raised, which might make this patch moot, fingers crossed.
* If we want to move forward with this, it might be good to apply for some quota increases on prod-3, to get the quotas in the same realm as prod-1 and prod-2. We will want 30 req/s instead of the current 20.
* In order to roll this out, we will need to drain some capacity from the current projects. That will be something along the lines of:
  ```
  bin/worker-deploy-gce drain production-1 worker-com --limit=9
  bin/worker-deploy-gce drain production-1 worker-org --limit=11
  bin/worker-deploy-gce decom production-1

  bin/worker-deploy-gce drain production-2 worker-com --limit=9
  bin/worker-deploy-gce drain production-2 worker-org --limit=11
  bin/worker-deploy-gce decom production-2
  ```

refs https://github.com/travis-ci/reliability/issues/217

### DIFF
--- a/gce-production-1/instance-counts.auto.tfvars
+++ b/gce-production-1/instance-counts.auto.tfvars
@@ -1,5 +1,5 @@
 {
-  "worker_managed_instance_count_com": 27,
-  "worker_managed_instance_count_org": 33,
+  "worker_managed_instance_count_com": 18,
+  "worker_managed_instance_count_org": 22,
   "worker_managed_instance_count_com_free": 0
 }
--- a/gce-production-2/instance-counts.auto.tfvars
+++ b/gce-production-2/instance-counts.auto.tfvars
@@ -1,5 +1,5 @@
 {
-  "worker_managed_instance_count_com": 27,
-  "worker_managed_instance_count_org": 33,
+  "worker_managed_instance_count_com": 18,
+  "worker_managed_instance_count_org": 22,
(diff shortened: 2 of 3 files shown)


gce-production-1/instance-counts.auto.tfvars
```diff
@@ -1,5 +1,5 @@
 {
-  "worker_managed_instance_count_com": 27,
-  "worker_managed_instance_count_org": 33,
+  "worker_managed_instance_count_com": 18,
```
Lyoness: not a blocker for merge, but more a curiosity on my part, how do we calculate the managed instance count (or pool size math also) for GCE?

igorwwwwwwwwwwwwwwwwwwww:
I took the previous value (e.g. 33 for org), multiplied by 2 because we have 2 projects currently, that gives a total of 66 instances for org. Then I divided by 3, since we want to now spread the same total number across 3 projects. That's where the new 22 instances per-project come from.

The goal is to keep the total number unchanged for now. Hope that explains the thinking!

PR #605, opened 2018-11-12
gce-production-3: resurrect, config

igorwwwwwwwwwwwwwwwwwwww:
This fixes our prod-3 config, which was outdated and failing due to disk size for windows jobs.

refs travis-ci/reliability#217


gce-production-3/worker.env
```diff
@@ -19,15 +19,20 @@ export TRAVIS_WORKER_GCE_PUBLIC_IP=false
 export TRAVIS_WORKER_GCE_PUBLIC_IP_CONNECT=false
 export TRAVIS_WORKER_GCE_RATE_LIMIT_PREFIX=worker-rate-limit-gce-api-3
 export TRAVIS_WORKER_GCE_RATE_LIMIT_TICK=2s
+export TRAVIS_WORKER_GCE_RATE_LIMIT_MAX_CALLS=30
```
igorwwwwwwwwwwwwwwwwwwww: This could be an issue, since afaik the quota there is still 20 req/s.

igorwwwwwwwwwwwwwwwwwwww: I've submitted a quota increase for this. 🤞

PR #611, opened 2018-11-15
Extract squignix stuff to module

meatballhat:
and stub in build-caching-production-1


soulshake:
I do have one question: what happens if the caching fails or becomes inaccessible for some reason? Is there a graceful fallback mode?

meatballhat:
@soulshake Thank you for the question! My thinking so far has been that proxy health detection could be part of the travis-build integration such that if the proxy is unavailable or sad that the apt conf override won't be written.

soulshake:
> proxy health detection could be part of the travis-build integration

Sounds good, we can continue the convo over there!

PR #614, opened 2018-11-19
Control build cache access via subnet + firewall rules

meatballhat:
Create a separate subnet for build-cache usage and explicitly allow internal traffic and health check ranges rather than allowing all public access to the instance ephemeral external IPs on port 80.


meatballhat:
This has already been `apply`'d, fwiw

PR #615, opened 2018-11-22
Increase conntracker threshold

joepvd:
Reduce the amount of networking errors due to hosts being temporarily blocked.

Increase the amount of allowed requests for a destination ip to get blocked.


Dirtybird57:
Help

Dirtybird57:
Help

PR #619, opened 2018-11-28
gce: local redis in gcp to reduce network latency

igorwwwwwwwwwwwwwwwwwwww:
As part of the investigation into https://github.com/travis-ci/worker/pull/564 we discovered what looks like a lot of time being spent on network latency when talking to redis.

Redis is in AWS `us-east-1`, whereas the workers are in GCP `us-central1`. The goal here is to make the redis instance local to the workers to make that network negligible.

This patch introduces a local redis instance per project. They cost about 50 bucks each, which is comparable to the 120 we are paying for the heroku-based one currently.

There are still some open questions about failovers, but those questions exist with our current setup too. :)

refs https://github.com/travis-ci/reliability/issues/222


igorwwwwwwwwwwwwwwwwwwww:
That's more like it:

screen shot 2018-11-28 at 17 08 21

PR #620, opened 2018-11-30
Modify build-cache resources based on feedback

meatballhat:


meatballhat:
This has already been applied in production :+1:

PR #621, opened 2018-12-03
Ensure ruby bits are checked

meatballhat:
so that syntax errors can't make it into master


.travis.yml
```diff
@@ -1,17 +1,18 @@
 language: bash
```
mjm: Is there a good reason to not change the language here to `ruby` and get the RVM install behavior for free?

meatballhat: That's fine by me, too!

mjm: Since it's already merged, I'd say it's up to you if you feel it's worth it.

PR #622, opened 2018-12-03
Get some ruby goodness for free :tada:

meatballhat:


Makefile
```diff
@@ -47,8 +47,7 @@ deps: $(DEPS)
 .PHONY: .ensure-rubocop
 .ensure-rubocop:
 	bundle version &>/dev/null || gem install bundler
-	bundle exec rubocop --version &>/dev/null || \
-		bundle install --jobs=3 --retry=3 --deployment --path $${BUNDLE_PATH:-vendor/bundle}
+	bundle exec rubocop --version &>/dev/null || bundle install
```
mjm: I'm not sure I understand what is happening with this change.

meatballhat: This means that iff someone runs `make deps` locally that they won't get the "deployement" style `bundle install`, which is more correct imho.

mjm: Ahh, okay cool!

PR #623, opened 2018-12-04
Add GKE cluster to Staging.

emdantrim:
Staging did not have GKE cluster.

Adding a GKE cluster.

## How can you test this?
GKE cluster now exists.

All of it!


gce-staging-1/k8s.tf
```diff
@@ -0,0 +1,46 @@
...
+  value = "${google_container_cluster.gke-staging-1.master_auth.0.client_certificate}"
+}
+
+output "client_key" {
+  value = "${google_container_cluster.gke-staging-1.master_auth.0.client_key}"
+}
+
+output "cluster_ca_certificate" {
+  value = "${google_container_cluster.gke-staging-1.master_auth.0.cluster_ca_certificate}"
+}
```
mjm: Can you add another output named `host` whose value is something similar to `https://10.182.64.100:6443` but with the name of the master instead of the IP?

I put a script at `bin/set-k8s-context` which uses the outputs that you've already included, as well as `host`, and creates a context in your local kubeconfig named after the graph directory.

The `macstadium-staging` and `macstadium-prod-1-cluster` directories also have this rule in their Makefile:

```
.PHONY: context
context:
	$(TOP)/bin/set-k8s-context $(ENV_NAME)
```

So you can run `make context` from a graph directory for a K8s cluster and then be able to connect to that cluster from then on with kubectl.

gce-staging-1/k8s.tf
```diff
@@ -0,0 +1,46 @@
...
+
+resource "random_id" "gke_password" {
+  byte_length = 32
+}
+
+resource "google_container_cluster" "gke-staging-1" {
+  name               = "gke-staging-1"
+  zone               = "us-central1-a"
+  initial_node_count = 3
+
```
mjm: What types of instances are the nodes in this cluster? Might be good to make it explicit, as I'm not sure what the default is.

My personal experience is that the nodes should probably be ~4GB RAM VMs. I'm running a personal cluster with 2GB and it is awfully tight on memory, even without much running on it.

modules/gke_cluster/gke.tf
```diff
@@ -0,0 +1,65 @@
...
+}
+
+variable "zone" {
+  default = "us-central1-a"
+}
+
+resource "google_container_cluster" "gke_cluster" {
+  name               = "${var.name}"
+  zone               = "${var.zone}"
+  min_master_version = "1.11"
```
emdantrim: TODO: variablize

PR #651, opened 2019-01-10
Move custom-2 worker on pod-2 to target .com

mjm:
This customer has finished moving their repos to .com, so we want to give their full capacity towards .com and stop targeting .org.


Lyoness:
Sweet.  Is there a twin PR to be done in kubernetes-config (I'm assuming?)

mjm:
@Lyoness There is not. The custom-2 worker in the macstadium-prod-1 cluster is already pointed at .com, and we don't have a cluster running in pod-2 yet, so this is the only relevant change we need to make for this. After this change, all custom-2 traffic will be going to .com.

PR #660, opened 2019-02-11
Integrate warmer into GCE jobs infrastructures

meatballhat:


bin/generate-latest-gce-images
```diff
@@ -4,7 +4,7 @@
 require 'json'
 
 DEFAULT_GCE_PROJECT_ID = 'eco-emissary-99515'
-DEFAULT_IMAGE_PREFIXES = 'bastion nat tfw'
+DEFAULT_IMAGE_PREFIXES = 'bastion'
```
Lyoness: I assume this is a change unrelated to to warmer and just changes made to NAT or `tfw`

gce-production-1/main.tf
```diff
@@ -170,9 +174,23 @@ resource "google_project_iam_member" "staging_1_workers" {
...
+}
+
 resource "google_project_iam_member" "production_2_workers" {
   count   = "${length(data.terraform_remote_state.production_2.workers_service_account_emails)}"
   project = "${var.project}"
   role    = "roles/compute.imageUser"
   member  = "serviceAccount:${element(data.terraform_remote_state.production_2.workers_service_account_emails, count.index)}"
 }
+
+resource "google_project_iam_member" "production_2_warmer" {
```
Lyoness: only other nit: why we name this member (and production workers) with `2` while staging  has`1`, given that we are in `gce-production-1`s main.tf?  Any significance?

Lyoness (approved):
Just curious questions in review.  LGTM!

bin/generate-latest-gce-images
meatballhat: I was getting an image lookup failure and realized that this list was longer than necessary :man_shrugging:

gce-production-1/main.tf
meatballhat: This resource only exists in `gce-production-1` and the purpose it to grant cloud images usage access to the _other_ `gce-.+` projects, which is why this resource has this name.

966556296635:
> _No description provided._

PR #661, opened 2019-02-26
[WIP] Add a Docker registry/pull-through cache

soulshake:

The purpose of this PR is to reduce egress for Docker Hub.

This PR:
- adds a GCE instance to serve as a Docker read-only registry/pull-through cache, and
- sets the `registry-mirrors` to the above instance on the Docker config of the default GCE build image

Addresses https://github.com/travis-ci/team-blue/issues/858


gce-staging-registry-1/main.tf
```diff
@@ -0,0 +1,56 @@
...
+}
+
+variable "gce_gcloud_zone" {}
+
+variable "index" {
+  default = 1
+}
+
+variable "project" {
+  default = "eco-emissary-99515"
```
meatballhat: I expect this should be `travis-staging-1`, yes?

gce-staging-registry-1/main.tf
```diff
@@ -0,0 +1,56 @@
...
+}
+
+variable "syslog_address_com" {}
+variable "syslog_address_org" {}
+
+variable "travisci_net_external_zone_id" {
+  default = "Z2RI61YP4UWSIO"
```
meatballhat: It's a trivial thing, but I would like to get away from referencing this directly and instead looking it up as with https://github.com/travis-ci/terraform-config/blob/9ee6c353d33d1589229c719db74f2a41761a6fe0/dns-production-0/main.tf#L45-L47

gce-staging-registry-1/main.tf
```diff
@@ -0,0 +1,56 @@
...
+    dynamodb_table = "travis-terraform-state"
+  }
+}
+
+provider "google" {
+  project = "${var.project}"
+  region  = "us-central1"
+}
+
+data "terraform_remote_state" "registry_staging_1" {
```
meatballhat: I don't see this being used, and I don't expect you will need to pull in the remote state of the same graph that is being built here.

modules/gce_docker_registry_cache/cloud-config.yml.tpl
```diff
@@ -0,0 +1,20 @@
...
+  encoding: b64
+  path: /etc/docker/registry/config.yml
+- content: '${base64encode(file("${here}/cloud-init.bash"))}'
+  encoding: b64
+  path: /var/lib/cloud/scripts/per-instance/99-travis-cloud-init
+  permissions: '0750'
+- content: '${base64encode(syslog_address)}'
+  encoding: b64
+  path: /var/tmp/travis-run.d/syslog-address
+- content: '${base64encode(file("${here}/daemon-direct-lvm.json"))}'
```
meatballhat: Is direct-lvm storage helpful in this case? My understanding was that all blob storage was external.

modules/gce_docker_registry_cache/cloud-init.bash
```diff
@@ -0,0 +1,29 @@
...
+  : "${VARTMP:=/var/tmp}"
+
+  export DEBIAN_FRONTEND=noninteractive
+
+  eval "$(tfw printenv docker)"
+  __setup_docker
+  __install_registry
+}
+
+__setup_docker() {
```
meatballhat: I think that this is now redundant/outdated given what `tfw` provides by default. See the squignix setup for a more recent example https://github.com/travis-ci/terraform-config/blob/master/modules/gce_squignix/cloud-init.bash

modules/gce_docker_registry_cache/main.tf
```diff
@@ -0,0 +1,125 @@
...
+}
+
+data "template_file" "cloud_config" {
+  template = "${file("${path.module}/cloud-config.yml.tpl")}"
+
+  vars {
+    here = "${path.module}"
+
+    docker_registry_config = <<EOF
+${file("${path.module}/registry-config.yml")}
```
meatballhat: With this as the body, I would move the `${file(...)}` directly into the template.

modules/gce_remote_docker/daemon-direct-lvm.json
```diff
@@ -17,5 +17,6 @@
   "tlsverify": true,
   "tlscacert": "/etc/docker/ca.pem",
   "tlscert": "/etc/docker/server-cert.pem",
-  "tlskey": "/etc/docker/server-key.pem"
+  "tlskey": "/etc/docker/server-key.pem",
+  "registry-mirrors": ["staging-1-docker-registry.gce-us-central1.travisci.net"]
```
meatballhat: I assume this should not be hardcoded :grin:  Maybe the generated config in `tfw` should account for registry mirrors: https://github.com/travis-ci/tfw/blob/bfbffd80e1ddbcfd13cf58b9fbe435949e810d3a/bin/tfw#L1270-L1288

soulshake: Ahhh I was looking all over for this. Thanks

modules/gce_docker_registry_cache/cloud-init.bash
soulshake: Where is the docker config being passed to the docker daemon?

soulshake:
Planning to just use [Container Registry's Docker Hub mirror](https://cloud.google.com/container-registry/docs/using-dockerhub-mirroring).

PR #662, opened 2019-03-19
[WIP] Create daemon.json to write to /etc/docker/daemon.json

soulshake:
This should be a better solution than https://github.com/travis-ci/terraform-config/pull/661.

Also need to revert https://github.com/travis-ci/tfw/pull/48


soulshake:
This won't work as gce_remote_docker is for testing.

PR #665, opened 2019-04-23
Improvements on GCE infra #863, instance group for gcleanup

DamianSzymanski:

With gcloud cleanup sitting on heroku we don't have access to redis on gpc, we want to move service to the instance on GPC

Build and deploy service to the instance group on GPC

any welcome


modules/gcloud_cleanup/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+variable "gcloud_cleanup_scale" {
+  default = "worker=1:Standard-1X"
+}
+
+variable "gcloud_cleanup_version" {
+  default = "master"
+}
+
+variable "gcloud_cleanup_docker_self_image" {
+  default = "travisci/gcloud-cleanup:latest"
```
soulshake: As far as I know, we're not yet making use of the `latest` tag, so [`travisci/gcloud-cleanup:latest` on Docker Hub](https://hub.docker.com/r/travisci/gcloud-cleanup/tags) is 6 months old.

modules/gcloud_cleanup/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+
+variable "gcloud_cleanup_opencensus_tracing_enabled" {
+  default = "false"
+}
+
+variable "gcloud_cleanup_scale" {
+  default = "worker=1:Standard-1X"
+}
+
+variable "gcloud_cleanup_version" {
```
soulshake: Where is this used?

DamianSzymanski: It's not used, will remove

modules/gcloud_cleanup/variables.tf
DamianSzymanski: Not sure what to put here, I used `b28c85d` for staging build, but I guess we want to use latest  here. We could push to latest after this merge https://github.com/travis-ci/gcloud-cleanup/pull/53 wdyt?

Duologic:
@DamianSzymanski Is this still relevant if we are running on GKE?

PR #666, opened 2019-05-20
Use google-beta on gc_instance_group_manager.net

Duologic:

This resolves an error when trying to (re-)create the bastion host in gce-staging-net-1:

> Error: module.gce_net.google_compute_instance_group_manager.nat[3]: "auto_healing_policies.0.initial_delay_sec": [REMOVED] This field is in beta. Use it in the the google-beta provider instead. See https://terraform.io/docs/providers/google/provider_versions.html for more details.

Introduce the `google-beta` provider, it works just like the `google` provider but you can set a `google-beta` on specific resource to use new features.

https://www.terraform.io/docs/providers/google/provider_versions.html#using-the-google-beta-provider

## How can you test this?

Executed against gce-staging-net-1.

Should we prevent using `google-beta`? If so, which solution do you propose?

### DIFF
--- a/gce-staging-net-1/main.tf
+++ b/gce-staging-net-1/main.tf
@@ -49,7 +49,7 @@ terraform {
   }
 }
 
-provider "google" {
+provider "google-beta" {
   project = "${var.project}"
   region  = "${var.region}"
 }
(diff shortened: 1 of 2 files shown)


Duologic:
Superseeded by #666

PR #667, opened 2019-05-20
Small fixes to make this repo easier to use

Duologic:

`make plan` returned changesets for gce_workers as user-data got changed, usually user-data is only changed for when setting up new machines not already running machines. IMO we should not have too much user-data scripts, I believe this can better be handled by configuration management.

The `worker_image` name is different then set on the machines itself, this syncs that to what is actually on GCE.

Also, hitch-hiking a fix for the Makefile as I like to keep this setup isolated from my own installation. I've put the terraform installation in a different location, this was partially handled by `$TF_INSTALLATION_PREFIX` but not yet reflected in the Makefile.

Last change, I've put a few extra dirs in `.gitignore` that showed up as changed but are actually runtime files for terraform or ruby.

These changes should limit the plan/diff output from terraform and make it generally easier to use it.

## How can you test this?

Run it against gce-staging-1 (plan/apply) and against gce-production-1 (plan).

Any feedback is welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,7 @@ graph.png
 local-development-0/local-development-0.tfstate
 macstadium-*/config/
 tfstate
+.cache/
+config/
+.bundle/
+vendor/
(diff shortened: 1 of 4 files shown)


Duologic:
Superseeded by https://github.com/travis-ci/terraform-config/pull/668

PR #672, opened 2019-06-25
Remove unused apps from the configuration

Duologic:

- gcloud-cleanup is running on kubernetes, no need to configure it with terraform anymore.
- warmer is not really being used right now, turning it of for now
- create a kubernetes cluster on prod-2, initially to host gcloud-cleanup
- order the nat_names differently on prod-1, see related commit f279f65

Remove configuration that is not actively used, less code = less bugs.

## How can you test this?

Run against staging.

If you know more about warmer, I'd like to hear about it.

### DIFF
--- a/bin/gce-export-workers
+++ b/bin/gce-export-workers
@@ -48,7 +48,6 @@ def main
   zones = options.fetch(:zones)
 
   command = %W[#{terraform} state rm]
-  command << "module.#{module_name}.heroku_app.gcloud_cleanup"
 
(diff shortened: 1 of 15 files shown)


Duologic:
I've worked from this branch to configure the current setup, prod-1 and prod-2 now match this configuration. Once build succeeds, I will merge this.

Duologic:
Just did some sanity checks, the NATs on prod-1 are not in line with the code yet, merging anyway.

PR #682, opened 2019-08-19
Increase NATs and max k8s nodes on prod-3

Duologic:
Our quota's have been increased, this will allow us to add more NAT machines and increase nodes/workers on prod-3

Increase the number of NATs: this should keep the existing IPs around.
NOTE: We should first create the new IPs (`terraform plan -target ...`) and communicate them

This also increases the max. nodes for GKE, the same as we have on the other projects.

## How can you test this?
This is tested module code and already running in prod.

Not  really.


Duologic:
Proposed change already on master.

PR #684, opened 2019-08-19
Create Services project

Duologic:
We want to run the services for Travis CI to run separate from the jobs/workers. This PR should provide a GCP project and Kubernetes cluster for this.

* Initialize/configure project from Terraform
* Use GCS for tfstate storage
* New networking module to not include the complicated NAT as on the workers.
* Refactored Kubernetes module.

## How can you test this?
Setup a staging project first.

Sure, feel free to discuss.


Duologic:
I was able to set this up on the staging project. Once billing is set up for production project, this can move forward.

Duologic:
https://github.com/travis-ci/terraform-config/pull/684/commits/30b6bdcd567e99fadbac3b822a91c025b7541a8c adds a target to create the Terraform state bucket, this should be run once before running Terraform.

PR #685, opened 2019-08-30
Allow cert manager webhook

DamianSzymanski:

Configure an additional firewall rule to allow the GKE control plane access to your webhook pod.
More info:
https://docs.cert-manager.io/en/latest/getting-started/webhook.html#running-on-private-gke-clusters

Add service cert-manager

## How can you test this?

Reproduce: Without rule by typing `kubectl api-resources` api returns 

> error: unable to retrieve the complete list of server APIs: admission.certmanager.k8s.io/v1beta1: the server is currently unable to handle the request

to apply rule `make plan && make apply` and verify `kubectl api-resources`

Welcome


Duologic (changes requested):
Maybe put it in `modules/gce_net_services/`?

travis-ci-prod-services-1/firewall.tf
```diff
@@ -0,0 +1,14 @@
...
+  project = "${module.project.project_id}"
+
+  source_ranges = ["172.16.0.0/28"]
+
+  allow {
+    protocol = "tcp"
+    ports    = ["6443"]
+  }
+
+  source_tags = ["services"]
```
Duologic: Does it need both `source_ranges` and `source_tags`? We could simply use `source_tags` and fill it in with a variable that is also used on the cluster's node pool. In case we then decide to change the tag, it changes on both locations at once.

Duologic (changes requested):

travis-ci-prod-services-1/firewall.tf
DamianSzymanski: source_ranges is needed

travis-ci-prod-services-1/modules.tf
```diff
@@ -7,7 +7,9 @@ module "project" {
 module "networking" {
   source = "../modules/gce_net_services"
 
-  project = "${module.project.project_id}"
+  project                  = "${module.project.project_id}"
+  cert_manager_enabled     = 1
+  cert_manager_source_tags = "${var.node_pool_tags}"
```
Duologic: See line 27 in this file, we should set the node_pool_tags on kubernetes_cluster dynamically too.

travis-ci-staging-services-1/modules.tf
```diff
@@ -7,7 +7,9 @@ module "project" {
 module "networking" {
   source = "../modules/gce_net_services"
 
-  project = "${module.project.project_id}"
+  project                  = "${module.project.project_id}"
+  cert_manager_enabled     = 1
+  cert_manager_source_tags = "${var.node_pool_tags}"
```
Duologic: See line 27 in this file, we should set the node_pool_tags on kubernetes_cluster dynamically too.

Duologic (changes requested):

